PR #13, opened 2014-10-03
Fix error handing in Tornado parser.

jmcarp:
The second argument to tornado.web.HTTPError, log_message, is expected
to be a string, but TornadoParser#handle_error was passing an Exception
object instead, which caused Tornado applications to time out. This
patch sends the message of the exception rather than the exception
itself, and adds a test case that also times out if the wrong argument
is passed to HTTPError.


sloria:
Thanks for the fix and also for bringing light to the fact that `TornadoParser#handle_error` was completely untested. =)

PR #15, opened 2014-10-12
Add optional source parameter to Arg.

jmcarp:
For use when values are keyed on variable-unsafe names, or when the developer just wants to rename keys:

``` python
args = {
    'content_length': Arg(source='Content-Length'),
    'query': Arg(source='q'),
}
```


sloria:
Thanks!

PR #16, opened 2014-10-29
Handle TypeError on parsing JSON.

jmcarp:
In Tornado, request.body is normally a string, but it can be a
concurrent.Future on streaming requests. This means we also have to
handle TypeErrors on parsing JSON.


sloria:
Thanks!

PR #24, opened 2014-11-30
Argument type validation with user-friendly error messages

Ch00k:


webargs/core.py
```diff
@@ -139,6 +141,21 @@ class Arg(object):
     .. versionchanged:: 0.5.0
         The ``use`` callable is called before type conversion.
     """
+    JSON_TYPES = {
```
sloria: I think `JSON_TYPES` can be defined at the module level rather than the class level. That way, if the user wants to modify the dictionary, they don't need to subclass `Arg`. This mapping should not differ between classes anyway.

webargs/core.py
```diff
@@ -139,6 +141,21 @@ class Arg(object):
...
+        'list': 'array',
```
sloria: Could the keys just be the type objects rather than the string names of the types? This would allow a user to included multiple classes with the same name in the mapping

``` python
from module_a import MyType as MyTypeA
from module_b import MyType as MyTypeB
JSON_TYPES = {
    MyTypeA: 'object',
    MyTypeB: 'object',
}
```

webargs/core.py
```diff
@@ -139,6 +141,21 @@ class Arg(object):
...
+        'int': 'integer',
+        'float': 'number',
+        'long': 'number',
+        'NoneType': 'null',
+        'dict': 'object',
+        'str': 'string',
+        'unicode': 'string',
+    }
+
+    NON_NULLABLE_TYPES = (list, tuple, dict, bool)
```
sloria: This can be a `set`.

webargs/core.py
```diff
@@ -163,15 +180,32 @@ def __repr__(self):
...
+            if self.type != noop:
+                msg = ('Expected type {} for {}, got {}'
+                       .format(self.JSON_TYPES[self.type.__name__], name,
+                               # TODO: Handle types not known by JSON_TYPES
+                               self.JSON_TYPES[type(ret).__name__]))
+
+                # Allow simple types (string, integer, number) to receive null
+                if ret is None and self.type in self.NON_NULLABLE_TYPES:
+                    raise ValidationError(self.error or msg)
+                if not isinstance(ret, self.type):
```
sloria: This isn't exactly correct. The `type_` argument specifies how a request value should be _converted_, not just validated. I think you will need to move this message-formatting logic to the `except` block below.

webargs/core.py
```diff
@@ -163,15 +180,32 @@ def __repr__(self):
...
         """Perform conversion and validation on ``value``."""
         ret = value
         for func in self.use_funcs:
             ret = func(ret)
-        try:
-            ret = self.type(ret)
-        except ValueError as error:
-            raise ValidationError(self.error or error)
+
+        if self.target == 'json':
```
sloria: I'm not sure that JSON requests need to be special-cased here. These error messages apply equally well to form bodies and querystring arguments.

sloria:
First pass code review finished. :+1:  

I think many of the tests are failing because you are not using `self.type` to do type conversion (you are only doing type checking against it). 

Please let me know if you need any help with this or have any questions. Thanks again for taking your time to do this.

webargs/core.py
Ch00k: While writing this I was actually trying to avoid conversion completely. I'd rather consider `type_` a type that a variable must be of when passed to the parser (this is especially important for JSON). If it's not of that type the parser fails, but not tries to convert.
My motivation is that we actually used conversion not for conversion per se but for validation - if we could not convert we fail. And I replaced it with just type checking.
Please let me know if I'm missing the point here.

webargs/core.py
Ch00k: This is arguable. So far I only had a chance to play with Flask and noticed that in all the other target except json body Flask always gets values as strings. Thus there probably is not reason to to even type checking for other targets.
Anyway, I need to more research to find out the best way to handle all out targets simultaneously.

webargs/core.py
Ch00k: I consider `JSON_TYPES` a constant, something that is not overridable. Basically the same as `TARGET_MAP`

sloria: I can see value in being able to modify the JSON types mapping. A user may want to support custom types, or use different nomenclature.

`TARGET_MAP` is not actually a constant--it can be modified via the `Parser.target_handler` decorator. I realize now that I made a naming convention error when making it all-caps. I will rename it to all-lowercase or `__target_map__`.

webargs/core.py
sloria: `type_` is analagous to werkzeug's [routing converters](http://werkzeug.pocoo.org/docs/0.9/routing/#builtin-converters); they convert input values to ensure that the view function receives arguments of the correct type.

If you only want to do type-checking of the raw input values, you should use the `validate` argument instead.

sloria: As you pointed out in another comment, form and querystring data come in as strings; it is useful to have the request parser convert these values for you.

Ch00k:
My understanding of `type_` was wrong from the very beginning. I changed the code a bit to actually do type conversion before validation.

Ch00k:
@sloria Ok, so the tests are no fine except of Python 2.6. I can't really find the reason it's failing though because I don't have Python 2.6 at hand. Can you please take a look?

webargs/core.py
```diff
@@ -163,15 +185,25 @@ def __repr__(self):
...
                 'required={self.required})>').format(self=self)
 
-    def _validate(self, value):
+    def _validate(self, name, value):
         """Perform conversion and validation on ``value``."""
         ret = value
         for func in self.use_funcs:
             ret = func(ret)
+
+        msg = 'Expected type {} for {}, got {}'\
```
sloria: In Py26, positional argument specifiers are required, so this must be

``` python
msg = 'Expected type {0} for {1}, got {2}'.format(...)
```

tests/test_core.py
```diff
@@ -233,11 +233,12 @@ def test_value_error_raised_if_invalid_target(request):
         p.parse_arg('foo', arg, request, targets=('invalidtarget', 'headers'))
     assert 'Invalid targets arguments: {0}'.format(['invalidtarget']) in str(excinfo)
 
-@mock.patch('webargs.core.Parser.parse_json')
-def test_conversion(parse_json, request):
-    parse_json.return_value = 42
-    arg = Arg(str)
-    assert Parser().parse_arg('foo', arg, request, targets=('json',)) == '42'
+# TODO: This is no longer valid. Leaving commented for now
```
sloria: This test should now be valid. Can you uncomment this?

Ch00k:
Tests are ok now, thanks! I'll be adding more tests to support my changes in the next couple of days.

webargs/core.py
```diff
@@ -111,6 +115,23 @@ def noop(x):
...
+    list: 'array',
+    tuple: 'array',
+    bool: 'boolean',
+    int: 'integer',
+    float: 'number',
+    long: 'number',
+    type(None): 'null',
+    dict: 'object',
+    str: 'string',
+    unicode: 'string',
```
sloria: In order to ensure support for binary and unicode strings on both Py2 and Py3, you could do the following:

``` python
# at the top of the file
if not PY2:
    # ...
    text_type = str
    binary_type = bytes
else:
    # ...
    text_type = unicode
    binary_type = str
```

``` python
JSON_TYPES = {
    text_type: 'string',
    binary_type: 'string',
    # ...
}
```

sloria:
Looking good so far!

Ch00k:
I added some tests. I think it's finally ready to be considered for a merge.

tests/test_core.py
```diff
@@ -32,54 +33,82 @@ def test_bad_validate_param(self):
...
         arg = Arg(type_=int)
-        assert arg.validated(42) == 42
-        with pytest.raises(ValidationError):
-            arg.validated('nonint')
+        assert arg.validated('foo', 42) == 42
+        with pytest.raises(ValidationError) as excinfo:
+            arg.validated('foo', 'nonint')
+        assert 'Expected type integer for foo, got string' in str(excinfo)
+
+    def test_validated_non_nullable_types(self):
```
sloria: Prefer pytest's parametrization functionality to `for` loops:

``` python
@pytest.mark.parametrize('arg_type',
[
    dict, list, tuple, bool
])
def test_validated_non_nullable_types(self, arg_type):
    arg = Arg(type_=arg_type)
    # ...
```

webargs/core.py
```diff
@@ -111,6 +115,22 @@ def noop(x):
     return x
 
 
+TYPES = {
+    list: 'array',
+    tuple: 'array',
```
sloria, line 120: Perhaps `set` should also be included here?

webargs/core.py
```diff
@@ -111,6 +115,22 @@ def noop(x):
...
+    int: 'integer',
+    float: 'number',
+    long_type: 'number',
+    type(None): 'null',
+    dict: 'object',
+    text_type: 'string',
+    binary_type: 'string',
+}
+
+NON_NULLABLE_TYPES = set([list, tuple, dict, bool])
```
sloria: Include `set` here as well?

tests/test_core.py
```diff
@@ -3,7 +3,8 @@
 
 import pytest
 
-from webargs.core import Parser, Arg, ValidationError, Missing, get_value, PY2
+from webargs.core import Parser, Arg, ValidationError, Missing, get_value, PY2, \
```
sloria: Use parentheses to break long lines:

``` python
from webargs.core import (
    Arg,
    get_value,
    Missing,
    Parser,
    PY2,
    ValidationError,
    #...
)
```

webargs/core.py
```diff
@@ -163,15 +184,24 @@ def __repr__(self):
...
                 'required={self.required})>').format(self=self)
 
-    def _validate(self, value):
+    def _validate(self, name, value):
         """Perform conversion and validation on ``value``."""
         ret = value
         for func in self.use_funcs:
             ret = func(ret)
+
+        msg = 'Expected type {0} for {1}, got {2}'\
```
sloria: Again, use parens to break long lines:

``` python
        msg = 'Expected type {0} for {1}, got {2}'.format(
            TYPES.get(self.type, self.type.__name__),
            name,
            TYPES.get(type(ret), type(ret).__name__)
        )
```

tests/test_core.py
```diff
@@ -32,54 +33,82 @@ def test_bad_validate_param(self):
...
+        assert arg.validated('foo', {}) == {}
+        assert arg.validated('foo', None) is None
+
+    def test_validated_text_type(self):
+        arg = Arg(type_=text_type)
+        assert arg.validated('foo', 42) == '42'
+
+    def test_validated_long_type(self):
+        arg = Arg(type_=long_type)
+        assert arg.validated('foo', 42) == 42
```
sloria, line 100: Could you also add a test case for the passing a `type_` argument that is not in the `TYPES` mapping?

sloria:
Looks good. :+1: Just a few minor changes and an added test case, and this should be good to merge.

tests/test_core.py
Ch00k: It's not working this way. `arg_type` is a `Mock` object and it does not have the `__name__` attribute thus `validated` method fails. Do I need to create a separate fixture for non_nullable_types list?

webargs/core.py
Ch00k, line 120: Added `set`.

tests/test_core.py
```diff
@@ -3,7 +3,20 @@
...
+    Missing,
+    get_value,
+    PY2,
+    text_type,
+    long_type,
+    NON_NULLABLE_TYPES,
+    TYPES
+)
+
+from collections import OrderedDict
```
sloria: `OrderedDict` is not available in the Py26 stdlib, so the build is failing. I believe the `uuid.UUID` type is consistent across Pythons. Perhaps you can test against that instead.

tests/test_core.py
Ch00k: It looks like we are overriding the builtin py.test request fixture, that's why it did not work work for me. I fixed it like this: https://github.com/Ch00k/webargs/pull/1/files Does it make sense?

sloria: Aha, nice catch. Yes, the fix you made makes sense. :+1:

Ch00k:
Looks like it's now ready. Please double-check.

webargs/core.py
```diff
@@ -111,6 +115,23 @@ def noop(x):
     return x
 
 
+TYPES = {
```
sloria: If we want this mapping to be configurable, it should not be named like a constant, i.e. uppercased. Perhaps lowercase `type_map`, or `__type_map__` (to be consistent with `Parser.__target_map__`.

Same thing with `NON_NULLABLE_TYPES`.

webargs/core.py
```diff
@@ -163,15 +185,25 @@ def __repr__(self):
...
                 'required={self.required})>').format(self=self)
 
-    def _validate(self, value):
+    def _validate(self, name, value):
         """Perform conversion and validation on ``value``."""
         ret = value
         for func in self.use_funcs:
             ret = func(ret)
+
+        msg = 'Expected type {0} for {1}, got {2}'.format(
```
sloria: Perhaps put double quotes around `{0}` and `{1}` to designate them as variants?

sloria:
Just a few more suggestions in the line comments. Once those are processed, this is good to merge.

webargs/core.py
Ch00k: What about the `Parser.DEFAULT_TARGETS`? What's the reason it's still a constant?

Ch00k:
Done.

sloria:
Thanks for your contribution!

PR #30, opened 2014-12-19
Handling Arg(type, multiple=True, allow_missing=True)

venuatu:
core.get_value gets the _Missing type, places it into a list and then fails the type conversion/validation.
The attached test gives a 400 with: `'message': 'Expected type "integer" for ids, got "_Missing"'` and only seems to be hit when there is something in the payload

A terribly quick hack is in: https://github.com/venuatu/webargs/commit/7824c968df7d2c7db9da7aba3d7f613a13fbd196 but it doesn't take allow_missing into consideration.
It looks like core.get_value might need a refactor to take allow_missing and/or required or pushing the function into the argument class.
What would be the best way of fixing this?

### DIFF
--- a/tests/test_flaskparser.py
+++ b/tests/test_flaskparser.py
@@ -329,6 +329,13 @@ def test_multiple_arg_allowed_missing(testapp):
         args = parser.parse(args)
         assert 'name' not in args
 
+def test_multiple_arg_allowed_missing_int_conversion(testapp):
+    args = {'ids': Arg(int, multiple=True, allow_missing=True)}
+    with testapp.test_request_context(path='/foo', method='post',
+            data=json.dumps({'fakedata': True}), content_type='application/json'):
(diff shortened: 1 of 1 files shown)


sloria:
@venuatu Thanks you for catching this and adding the failing test. I think the solution here is to allow `Missing` to be returned by `get_value`, even when `multiple=True`.

``` python
def get_value(d, name, multiple):
    val = d.get(name, Missing)
    if multiple and val is not Missing:
        #...
    return val
```

That way, the `Missing` value can be handled by the `Parser`, and `required` and `allow_missing` should behave as expected. 

Will need to think more about possible edge cases, but I think this is the desired solution since it is a minimal change doesn't affect the public interface.

sloria:
Confirmed that the above fix works. Merged into `dev`. Thanks again for reporting!

PR #35, opened 2015-03-19
Store the argument name inside the exception object.

### LINKED ISSUE #34: Include failing argument name for validation errors
When a validation occurs I want to be able to return the name of the failing argument. At the moment, this is not possible since the validation error contains just a string. An example response:

``` json
{
    "message": "A validation error occured.",
    "params": {
        "foo": "Longer than 10"
    }
}
```

Is this something you feel is worth adding?

stas:
This should fix #34 

I can see @alexmic suggested we should store also the value of the argument.
My opinion is that this is not a good idea. Especially if we want to validate files or payloads of bigger sizes (consider a dict, base64 hash, etc...)

Let me know if you think that won't be an issue.

Thanks in advance for reviewing this.

### DIFF
--- a/webargs/core.py
+++ b/webargs/core.py
@@ -40,10 +40,10 @@ class ValidationError(WebargsError):
 
         Store status_code and additonal data.
     """
-    def __init__(self, error, status_code=400, **data):
+    def __init__(self, error, status_code=400, arg_name=None, **data):
         self.message = text_type(error)
         self.status_code = status_code
-        self.arg_name = data.pop('arg_name') if 'arg_name' in data else None
+        self.arg_name = arg_name
         self.data = data
         super(ValidationError, self).__init__(self.message)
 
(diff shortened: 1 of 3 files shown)


alexmic:
Hey @stats, my suggestion was just for the name. "Longer than 10" was the validation error message.

stas:
@alexmic sorry, my bad.

alexmic:
No worries – thanks for taking a look at this, much appreciated.

stas:
Looks like travis is broken, might need a pip version update.

webargs/core.py
```diff
@@ -43,6 +43,7 @@ class ValidationError(WebargsError):
     def __init__(self, error, status_code=400, **data):
```
sloria: Why not have `arg_name` as a named argument instead of pulling of `data`?

sloria:
Thanks @stas for your contribution. If you pull from `dev` the Travis build should be fixed.

This looks good; just a minor change suggested in my comment. Otherwise, this should be good to merge.

stas:
I updated the PR based on your suggestion @sloria. Thanks.

This should be good to merge.

sloria:
Thanks @stas !

PR #36, opened 2015-03-23
Support for webapp2 on Google App Engine

### LINKED ISSUE #6: support for webapp2 framework?
Hi, just wondering if you planned on supporting the webapp2 framework which is using standard webob requests objects. This would make this library usable on appengine with more frameworks and would be awesome.
Thanks!

hello-josh:
I've also added an example and something to the docs.

Resolves #6

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -16,6 +16,7 @@ Django>=1.6.5
 bottle>=0.12.3
 tornado>=4.0
 pyramid>=1.5.2
+webapp2>=2.5.2
 
 # Syntax checking
 flake8==2.4.0
--- a/tests/test_webapp2parser.py
+++ b/tests/test_webapp2parser.py
@@ -1,0 +1,124 @@
+# -*- coding: utf-8 -*-
+"""Tests for the webapp2 parser"""
+import urllib
+import json
+
+import pytest
+from webargs.core import PY2
+
+pytestmark = pytest.mark.skipif(not PY2, reason='webapp2 is only compatible with Python 2')
+
+import webtest
+
+if PY2:
+    import webapp2
+
+from webargs import Arg, ValidationError
+from webargs import webapp2parser
+
+parser = webapp2parser.Webapp2Parser()
+
+hello_args = {
+    'name': Arg(unicode, default='World'),
+}
+
+hello_multiple = {
+    'name': Arg(multiple=True),
+}
+
+hello_validate = {
+    'num': Arg(int, validate=lambda n: n != 3, error="Houston, we've had a problem."),
+}
+
+
+def test_parse_querystring_args():
+    request = webapp2.Request.blank('/echo?name=Fred')
+    assert parser.parse(hello_args, req=request) == {'name': 'Fred'}
+
+
+def test_parse_querystring_multiple():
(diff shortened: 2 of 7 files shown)


hello-josh:
The webapp2 stuff only runs on python 2.7, I may need to update the tests so they only execute in the python 2.7 env for Travis

.gitignore
```diff
@@ -46,3 +46,4 @@ README.html
 
 temp/*
 .konchrc
+.idea
```
sloria: Files/folders like this belong in a global gitignore: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

dev-requirements.txt
```diff
@@ -16,6 +16,7 @@ Django>=1.6.5
 bottle>=0.12.3
 tornado>=4.0
 pyramid>=1.5.2
+webapp2==2.5.2
```
sloria: We haven't been pinning to versions of frameworks so that Travis CI will always test against the latest released version. For consistency, we should do the same with webapp2.

tests/test_webapp2parser.py
```diff
@@ -0,0 +1,121 @@
...
+hello_multiple = {
+    'name': Arg(multiple=True),
+}
+
+hello_validate = {
+    'num': Arg(int, validate=lambda n: n != 3, error="Houston, we've had a problem."),
+}
+
+
+def _assert_equals(result, expected):
```
sloria: What is the purpose of this function?

tests/test_webapp2parser.py
```diff
@@ -0,0 +1,121 @@
...
+import webapp2
+import urllib
+import json
+
+import pytest
+import webtest
+from webargs import Arg, ValidationError
+from webargs import webapp2parser
+
+parser = webapp2parser.Webapp2Parser()
```
sloria: To skip this test file on Python 3, you can do:

``` python
from webargs.core import PY2
pytestmark = pytest.mark.skipif(not PY2, reason='webapp2 is only compatible with Python 2')
```

sloria:
Thanks @Trii ! Just a few comments to address and this should be good to merge.

hello-josh:
Thanks for the feedback. I'll address these, rebase it into a single commit, and push it back up shortly.

dev-requirements.txt
hello-josh: Google app engine pins webapp2 to specific versions. I did the same here just in case they release a new version to pypi that doesn't match the available prod environment.

hello-josh:
ugh, importing webapp2 causes py3 syntax errors. I need to http://rtfd.org/projects/pytest/ about the py.test way to skip the import since I usually use nose and unittest.

hello-josh:
All fixed finally. Would you like me to combine these into a single commit?

sloria:
Thanks @Trii . No need to squash; I prefer the provenance of multiple commits.

I will give this a closer look when I'm home from work this evening.

hello-josh:
Great! On a side-note, I am on a similar path as your smore project. I am trying to integrate swagger doc generation, marshaling/unmarshaling, as well as validation based on the swagger schema into webapp2 right now so this helps a lot!

PR #37, opened 2015-03-24
Consolidate the error behaviour for required args.

stas:
This adds the argument name to the validation exception.
Required args errors can now be overridden, thus following the
existing argument API.

Some personal notes...
Current error messages are spread all-over the codebase.
I suggest adding the messages to the validation exception class
as class attributes. This should provide an easier way to customize
general error messages and make it easier to refactor/test things.

Thanks in advance for reviewing the changes.

### DIFF
--- a/webargs/core.py
+++ b/webargs/core.py
@@ -81,6 +81,23 @@ def _ensure_list_of_callables(obj):
     return validators
 
 
+def _raise_required(arg, arg_name):
+    """Raises an exception for a missing required argument.
+
+    If the argument required attribute carries a message, it will be used
+    as the exception message.
+
+    TODO: Get rid of it by DRY-ing up the nested args parsing.
+
+    :raises: RequiredArgMissingError
+    """
+    if isinstance(arg.required, basestring):
+        msg = arg.required
+    else:
+        msg = 'Required parameter "{0}" not found.'.format(arg_name)
+    raise RequiredArgMissingError(msg, arg_name=arg_name)
+
+
 class _Missing(object):
     """Represents a value that is missing from the set of passed in request
     arguments.
@@ -247,11 +264,7 @@ class Arg(object):
                     val = ret[key]
                 except KeyError:
                     if nested_arg.required:
-                        if isinstance(nested_arg.required, basestring):
-                            msg = nested_arg.required
-                        else:
-                            msg = 'Required parameter "{0}" not found.'.format(key)
-                        raise RequiredArgMissingError(msg, arg_name=key)
+                        _raise_required(nested_arg, key)
                 else:
                     ret[key] = nested_arg.validated(key, val)
         return ret
@@ -367,11 +380,7 @@ class Parser(object):
                 else:
                     value = argobj.default
             if argobj.required:
-                if isinstance(argobj.required, basestring):
-                    msg = argobj.required
-                else:
-                    msg = 'Required parameter "{0}" not found.'.format(name)
-                raise RequiredArgMissingError(msg, arg_name=name)
+                _raise_required(argobj, name)
         return value
 
     def parse(self, argmap, req, locations=None, validate=None, force_all=False):
--- a/webargs/core.py
+++ b/webargs/core.py
@@ -14,6 +14,7 @@ if not PY2:
     text_type = str
     binary_type = bytes
     long_type = float
+    basestring = (str, bytes)
 else:
     iteritems = lambda d: d.iteritems()
     text_type = unicode  # noqa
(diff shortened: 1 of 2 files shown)


sloria:
Thanks @stas. I will look at this more closely within the next day or two.

sloria:
It may make sense to use a similar API to marshmallow's (see https://github.com/marshmallow-code/marshmallow/issues/78) whereby the required validation message can be different from the "user" validators' message.

It would look like:

``` python
args = {
    'username': Arg(str, required='Must provide username', 
        validate=lambda v: len(v) > 6, error='Must be > 6 characters')
}
```

> Current error messages are spread all-over the codebase.
> I suggest adding the messages to the validation exception class
> as class attributes. 

Yes, I think that would be an improvement. Would you like to send that change in a separate PR?

stas:
> It may make sense to use a similar API to marshmallow's (see marshmallow-code/marshmallow#78) whereby the required validation message can be different from the "user" validators' message.

Alright, this sounds good.

> Yes, I think that would be an improvement. Would you like to send that change in a separate PR?

Yep, I'll try to make some time. Worst case, I will add an issue with a reference to this discussion.
Thanks for all the help with this!

webargs/core.py
```diff
@@ -149,8 +149,11 @@ class Arg(object):
         If ``None``, no type conversion will be performed.
     :param default: Default value for the argument. Used if the value is not found
         on the request. May be a callable.
-    :param bool required: If ``True``, the :meth:`Parser.handle_error` method will be
-        invoked if this argument is missing from the request.
+    :param required: If passes truth value test, the :meth:`Parser.handle_error`
+        method will be invoked if this argument is missing from the request.
+        Set a custome error message to be used instead of the default one.
```
sloria: "custom"

webargs/core.py
```diff
@@ -149,8 +149,11 @@ class Arg(object):
...
+        This is backwards compatible with the ``error`` value, but it is subject
```
sloria: I'm not clear on the purpose of this message.

webargs/core.py
```diff
@@ -245,9 +248,13 @@ def _validate(self, name, value):
                     val = ret[key]
                 except KeyError:
                     if nested_arg.required:
-                        raise RequiredArgMissingError(
-                            'Required parameter "{0}" not found.'.format(key)
-                        )
+                        req_type = (nested_arg.required).__class__.__name__
+                        if req_type in ['str', 'unicode']:
```
sloria: Why not `if isinstance(nested_arg.required, basestring)` (you would have to add `basestring = (str, bytes)` up top for py3 compat)?

webargs/core.py
```diff
@@ -363,9 +370,13 @@ def parse_arg(self, name, argobj, req, locations=None):
                 else:
                     value = argobj.default
             if argobj.required:
-                raise RequiredArgMissingError(
-                    'Required parameter "{0}" not found.'.format(name), arg_name=name
-                )
+                req_type = (argobj.required).__class__.__name__
```
sloria: Could we DRY this up with a helper function or method? Something like `_raise_required(arg)` ?

sloria:
Pass finished. :+1: Some minor refactoring and this should be good to merge.

webargs/core.py
stas: Yeah, I wasn't sure what's the best way to go about this. I've seen other people using the `six` package for this purpose, which obviously is not a solution...

I'll give it a try using `basestring`.

webargs/core.py
stas: I'll take a look at this, the problem with `RequiredArgMissingError` is that the same use-case gets handled in two places.

stas:
@sloria, could you take another look.

I didn't want to move the required argument message handling in a separate method, instead I tried to find a way to avoid handling required args in two places. To my despair I think there's something wrong with the way nested args are being handled now, I would say it is a little _smelly_.

This looks like a chicken and the egg problem. I believe the parser should be responsible for handling all the missing required args errors and leave args outside of this.

I would suggest leaving this the way it is now (or move the error extraction part right inside the  `RequiredArgMissingError` constructor) and add a todo task for future, describing some ideas how this can be improved.

Please let me know what you think.
And thanks for all the feedback so far.

sloria:
I agree that the nested arg handling could use some work, but that's for another PR. =)

In the meantime, I still think you could reduce the duplication with a simple helper function. Something like:

``` python
def _raise_required(arg, arg_name):
    if isinstance(arg.required, basestring):
        msg = arg.required
    else:
        msg = 'Required parameter "{0}" not found.'.format(arg_name)
    raise RequiredArgMissingError(msg, arg_name=arg_name)
```

stas:
@sloria last review please :)

sloria:
Thanks @stas!

PR #43, opened 2015-05-20
Fixes #42 Pyramid parser use_kwargs exception when used.

### LINKED ISSUE #42: Pyramid parser use_kwargs throws exception when used
The following code using the pyramid parser throws an exception:

``` python
@parser.use_kwargs({'myvalue': Arg(int)})
def baz(request, myvalue):
    return {'myvalue': myvalue}
```

The exception:

```
        kwargs['as_kwargs'] = True
>       return self.use_args(*args, **kwargs)
E       TypeError: use_args() got an unexpected keyword argument 'as_kwargs'
```

hartror:
Resolves #42.

### DIFF
--- a/tests/test_pyramidparser.py
+++ b/tests/test_pyramidparser.py
@@ -61,6 +61,10 @@ def testapp():
         def __call__(self, args):
             return args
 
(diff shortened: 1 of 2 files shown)


sloria:
Thanks!

PR #45, opened 2015-05-20
Use the parser's locations in `use_args`

### LINKED ISSUE #44: `use_args` does not use the parser's locations
If I specify custom locations in a parser:

``` python
parser = FlaskParser(locations=('data', 'query'))

@parser.location_handler('data')
def parse_data(req, name, arg):
    ...
```

Then decorate a method with `use_args`:

``` python
class WidgetList(MethodView):

    @parser.use_args(widget_args)
    def post(self, args):
    ...
```

Those custom locations are ignored and the default locations are used:

https://github.com/sloria/webargs/blob/f57c9b730d9b274edce85de38cdc39fdbdc2ecca/webargs/core.py#L445

This requires me to override locations in every `use_args` decorator:

``` python
    #                                v
    @parser.use_args(my_widget_args, locations=parser.locations)

```

Was this intentional? To me, it seems like line 445 in `use_args` should be:

``` python
locations = locations or self.locations
```

jacebrowning:
Fixes https://github.com/sloria/webargs/issues/44

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -636,6 +636,21 @@ def test_use_args(web_request, parser):
         return args
     assert viewfunc() == {'username': 'foo', 'password': 'bar'}
 
+def test_use_args_with_custom_locations_in_parser(web_request, parser):
+    custom_args = {
+        'foo': Arg(str),
+    }
+    web_request.json = {}
+    parser.locations = ('custom',)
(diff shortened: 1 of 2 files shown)


jacebrowning:
@sloria Thanks for merging! Any rough ballpark on a `0.13.1` release?

PR #47, opened 2015-06-06
Pyramid request matchdict support

hartror:
The Pyramid request matchdict contains parsed patterns from the url. Adding this functionality does a few useful things:
1. Type coercion.
2. Advanced validation.
3. Clearly marking a view as expecting a route with a particular matchdict value.

I don't love how I've done this:

``` #!python
__location_map__ = {
     'matchdict': 'parse_matchdict'
}
 __location_map__.update(core.Parser.__location_map__)
```

However not sure what else to do that would also have that level of clarity.

### DIFF
--- a/webargs/pyramidparser.py
+++ b/webargs/pyramidparser.py
@@ -39,6 +39,12 @@ logger = logging.getLogger(__name__)
 class PyramidParser(core.Parser):
     """Pyramid request argument parser."""
 
+    __location_map__ = {
+        'matchdict': 'parse_matchdict'
+    }
+
+    __location_map__.update(core.Parser.__location_map__)
+
     def parse_querystring(self, req, name, arg):
         """Pull a querystring value from the request."""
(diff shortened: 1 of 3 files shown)


webargs/pyramidparser.py
```diff
@@ -39,6 +39,12 @@ def hello_world(request, args):
 class PyramidParser(core.Parser):
     """Pyramid request argument parser."""
 
+    __location_map__ = {
+        'matchdict': 'parse_matchdict'
+    }
+
+    __location_map__.update(core.Parser.__location_map__)
```
sloria: One alternative that might be slightly cleaner:

``` python
__location_map__ = dict(matchdict='parse_matchdict', **core.Parser.__location_map__)
```

sloria:
Looks good. Thanks @hartror. Just one suggested change, and this should be good to merge.

hartror:
Done @sloria.

Issue #46, opened 2015-05-29
default not working with nested args


Lemmons:
Nested args aren't respecting the default keyword.

For example:

``` python
@use_args({
    "param1": Arg(str, required=True, target="json"),
    "param2": Arg(str, required=False, default='bar', target="json"),
})
```

Behaves as expected. When making a request with the payload`{"param1": "foo"}`, `param2` is correctly added with the value `"bar"`.

On the other hand, with this:

``` python
@use_args({
    "param": Arg({
        "sub_param1": Arg(str, required=True, target="json"),
        "sub_param2": Arg(str, required=False, default='bar', target="json"),
    }),
})
```

making a request `{"param": {"sub_param1": "foo"} }` does not add `sub_param2`.

Without looking to the code too much, my guess would be this has the same root cause as https://github.com/sloria/webargs/issues/39 and https://github.com/sloria/webargs/issues/40.

PR #49, opened 2015-06-18
Added support for nested defaults and dests.

### LINKED ISSUE #40: source, dest arguments not working with nested args
Nested arg parsing does not respect source, dest arguments.

stas:
This should fix #40 and #46 

@sloria please take a look and let me know if this needs any further improvements.

Thanks in advance.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -619,6 +619,24 @@ def test_parse_with_dest(web_request):
     parsed = parser.parse(args, web_request, locations=('json',))
     assert parsed == {'content_type': 'application/json'}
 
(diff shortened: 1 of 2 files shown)


sloria:
This is a fine fix for the time being, but as you probably know, it duplicates much of the Parser's parsing logic. We will defer a cleaner solution for now.

sloria:
Thanks for the fix!

stas:
Yep, thanks for merging this even though there's a lot more work that needs to be done.

I'd like to help, but unfortunately not right now. Feel free to ping me if you need a second opinion on future related pull requests.

PR #52, opened 2015-07-01
Simplified nested destination parsing.

stas:
Please take a look at this.
This now covers the situation where the key is totally missing from payload.

We found this issue internally, sorry it got away in the #49 

Thanks in advance.


sloria:
Good catch!

PR #53, opened 2015-07-16
fix the bug, when get null in json

jayzh1010:
if the request type is json and has null type, the value will be a string 'None', but not real None,


sloria:
This is now fixed in the new marshmallow-based webargs

``` python
parser.parse({'foo': fields.Str(allow_none=True)}, request) == {'foo': None}
```

Issue #54, opened 2015-08-13
Null value handling


marcellarius:
Null values in webargs are currently not handled very well.
- None must not be passed to the type function.
  Python's builtin type functions don't handle None nicely (e.g. `str(None) == 'None'`)
- There needs to be a way to reject null values where they are not valid.
- required=True currently accepts null values, which can be a bit surprising.

I think the simplest solution is to skip type conversion for None values and rely on
the existing validation feature to implement a not-null check.  e.g. 

``` python
Arg(str, required=True, validate=lambda x: x is not None)
```

The documentation for the 'required' parameter should make it clear that None values are not 
considered missing.

Thoughts?

PR #55, opened 2015-08-17
Specify Mock version in dev-requirements

marcellarius:
The latest version of Mock (1.3.0) in PyPI breaks some existing test cases.  Specifying version 1.0.1 to keep things working.

PR #56, opened 2015-08-17
No null type conversion

marcellarius:
Issue #54 

Don't call the type conversion function if the parsed value is None.
I've also summarised the order that argument validation takes place in the Arg class docstring.


marcellarius:
The test failures are addressed in #55, due to a new version of Mock breaking existing tests.

sloria:
Indeed, this looks like the desired behavior. Thanks @marcellarius !

PR #57, opened 2015-08-17
Option to treat null as missing 

marcellarius:
In some situations it may be useful to treat null values in your JSON as equivalent to a missing value.

This adds a `none_as_missing` parameter to the `Arg` constructor, which when set to True will cause the `Parser.parse_arg` function to replace a `None` value with `Missing`.

This addresses the same issue as pull request #53 but shouldn't impact existing code.

A hypothetical alternative could be implemented with the `use` functions; however this currently couldn't work without shifting around some of the validation code.

``` python
def none_as_missing(value):
    return webargs.core.Missing if value is None else value

args = {
    "name": Arg(required=True, type=str, use=none_as_missing, validate=not_empty)
}
```

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -411,6 +411,26 @@ def test_value_error_raised_if_invalid_location(web_request):
         p.parse_arg('foo', arg, web_request, locations=('invalidlocation', 'headers'))
     assert 'Invalid locations arguments: {0}'.format(['invalidlocation']) in str(excinfo)
 
+def test_none_as_missing_sets_default(parser, web_request):
+    web_request.json = {"name": None}
+    args = {"name": Arg(none_as_missing=True, default="Bob")}
+    result = parser.parse(args, web_request, locations=('json',))
+    assert result["name"] == "Bob"
+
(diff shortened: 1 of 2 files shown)


sloria:
It may be worth generalizing this solution to treat other values as missing. I could see a use case for treating empty strings or empty Arrays as missing values.

sloria:
Closing this for now. 

With the new marshmallow-based fields, I recommend something like the following pattern:

``` python

class Skippable(fields.Field):

    def __init__(self, child_cls, *args, **kwargs):
        kwargs['allow_none'] = True
        self.child = child_cls(*args, **kwargs)
        super().__init__(*args, **kwargs)

    def deserialize(self, value, attr=None, data=None):
        if value is None:
            return missing
        return self.child.deserialize(value, attr, data)

    def _serialize(self, *args, **kwargs):
        return self.child._serialize(*args, **kwargs)

hello_args = {
    'name': Skippable(fields.Str)
}
@use_args(hello_args)
def index(args):
    name = args.get('name', 'World')
    return jsonify({'message': 'Welcome, {}!'.format(name)})
```

Issue #61, opened 2015-09-20
Use marshmallow under the hood


sloria:
### Proposal

Replace `Arg` objects with marshmallow fields. Use marshmallow for validation.
### Why?

Webargs and marshmallow have redundant functionality for validating input data against a user-defined schema. Marshmallow does a better job of than webargs in a number of respects:
- Nested validation is much more powerful in marshmallow. marshmallow's `Nested` field would solve #39 and #51.
- Marshmallow does error bundling. Solves #58 
- Marshmallow handles null and missing inputs in a more predictable way.
- Marshmallow has more consistent error messaging.
- Marshmallow's validation is more stable and well-tested
### Proposed API

``` python
from webargs import use_kwargs, fields


@use_kwargs({
    'title': fields.Str(required=True, location='json'),
    'description': fields.Str(required=False),
})
def myview(title, description):
    return 'title: {}, description: {}'.format(title, description)


# with nesting
@use_kwargs({
    'data': fields.Nested({
        'title': fields.Str(),
        'description': fields.Str(),
    })
})
def myview(data):
    return 'title: {}, description: {}'.format(**data)


# respects load_from
@use_kwargs({
    'x_custom_header': fields.Str(load_from='X-Custom-Header')
})
def myview(x_custom_header):
    return 'Header: {}'.format(x_custom_header)


# Passing in a Schema
from marshmallow import Schema

class PostSchema(Schema):
    title = fields.Str(required=True, location='json')
    description = fields.Str(required=False)
    id = fields.Int(dump_only=True)  # won't be included in parsed output

@use_kwargs(PostSchema)
def myview(title, description):
    return 'title: {}, description: {}'.format(title, description)
```

marcellarius:
This seems like a good solution to me.

hello-josh:
+1

sloria:
Making lots of progress on this.

I haven't yet decided where fields should be imported from.

Here are the alternatives (note that `Nested` is a webargs-specific field):

(1) Import from `webargs.fields`

``` python
from webargs import fields
from webargs.flaskparser import use_args
from marshmallow import validate

@use_args({
    'data': fields.Nested({
        'id': fields.Int(),
        'email': fields.Str(validate=validate.Email()),
    })
})
```

(2) Import from `webargs`

``` python

import webargs as wa
from webargs.flaskparser import use_args
from marshmallow import validate

@use_args({
    'data': wa.Nested({
        'id': wa.Int(),
[...]

PR #62, opened 2015-09-21
Use marshmallow to define args and do validation

sloria:
Implements #61 .

Todo:
- [x] Replace Arg with marshmallow fields
- [x] Port framework-specific modules
  - [x] Flask
  - [x] Django
  - [x] Bottle
  - [x] Tornado
  - [x] webapp2
  - [x] Pyramid
- [x] Add webargs.fields ? 
- [x] ~~Allow field validators to return a custom http status code ?~~ (deferred)
- [x] Pass Schemas to parser
- [x] Update documentation


jmcarp:
:tada:

PR #64, opened 2015-09-28
Fix typo "message" to "messages"

hyunchel:
In the documentation for flask framework support, "message" key is expected
from "data" variable. It turns out, however, "data" contains a plural form
"messages" as the key.
For the purpose of this PR, code base is honored over the documentation.


sloria:
Thanks!

sloria:
Actually, I realized there are other things wrong with that example.
- Webargs returns a 422 response by default--not 400.
- The preferred way to access messages is to access `err.data['exc'].messages`. 

I will fix these now.

hyunchel:
Oh I should've caught those...sorry about that.
I'll make a better PR next time!

Issue #66, opened 2015-09-29
Support delimited list parameters


jmcarp:
Currently, webargs supports list parameters in query strings using repeated parameters: `foo=bar&foo=baz`. But it's also a common pattern to encode lists in query strings using delimited values: `foo=bar,baz`, `foo=bar|baz`, etc. This is described using the `collectionFormat` in swagger: https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#fixed-fields-8. It would be helpful to expose control over the formatting of list values to support both repeated and delimited parameters.

Here are a few thoughts about interfaces:
- Use a subclass of `List`, like `DelimitedList`, that splits and joins on a delimiter on de/serializing.
- Add an optional `delimiter` argument to `List` that has the same behavior as above.
- Add an option to `Schema.Meta` that controls de/serialization of lists:

``` python
class ArgSchema(Schema):
    class Meta:
        list_delimiter = ','
```

Not sure if I like any of these. What do you think @sloria?

sloria:
The 1st option seems best. `List` is not meant to deserialize from strings, so that rules out 2. Option 3 doesn't seem ideal because you wouldn't be able to use the feature with dictionary argmaps.

jmcarp:
Agreed with everything you said, but the first option strikes me as a little underpowered. In most use cases, all list parameters for a given resource will be encoded in the same way (all repeated parameters, or all delimited parameters), so I still like a higher-level option--what about an argument passed to `parse` and `use_(kw)args`? That said, I still think the lower-level option is independently useful, so I'm down to take a pass at it.

By the way, I don't particularly like delimited query parameters, but they seem to be pretty commonly used, and are even recommended by JSON-API, which is why I'd like to add support.

sloria:
Good point. I want to avoid adding a custom `Schema` class in webargs for the time being. Would it be enough if the delimiter option were configurable at the Field class level _and_ as a constructor argument?

PR #67, opened 2015-09-30
Delimited lists

jmcarp:
Proposed implementation of #66. Interface:

``` python
args = {'values': fields.DelimitedList(fields.Int())}
args = {'values': fields.DelimitedList(fields.Int(), delimiter='|')}
...
```

### DIFF
--- a/webargs/fields.py
+++ b/webargs/fields.py
@@ -43,19 +43,19 @@ class DelimitedList(ma.fields.List):
 
     :param Field cls_or_instance: A field class or instance.
     :param str delimiter: Delimiter between values.
-    :param bool dump_string: Dump values to string.
+    :param bool as_string: Dump values to string.
     """
 
     delimiter = ','
 
-    def __init__(self, cls_or_instance, delimiter=None, dump_string=False, **kwargs):
+    def __init__(self, cls_or_instance, delimiter=None, as_string=False, **kwargs):
         self.delimiter = delimiter or self.delimiter
-        self.dump_string = dump_string
+        self.as_string = as_string
         super(DelimitedList, self).__init__(cls_or_instance, **kwargs)
 
     def _serialize(self, value, attr, obj):
         ret = super(DelimitedList, self)._serialize(value, attr, obj)
-        if self.dump_string:
+        if self.as_string:
             return self.delimiter.join(format(each) for each in value)
         return ret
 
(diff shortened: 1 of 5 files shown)


webargs/fields.py
```diff
@@ -36,3 +36,19 @@ def __init__(self, nested, *args, **kwargs):
...
+
+    delimiter = ','
+
+    def __init__(self, cls_or_instance, delimiter=None, **kwargs):
+        self.delimiter = delimiter or self.delimiter
+        super(DelimitedList, self).__init__(cls_or_instance, **kwargs)
+
+    def _serialize(self, value, attr, obj):
+        ret = super(DelimitedList, self)._serialize(value, attr, obj)
+        return self.delimiter.join(ret)
```
sloria: I'm not sure this should serialize to a string by default, In the typical use case--rendering an HTTP response with serialized data--wouldn't you usually want a list?

In any case, I think a parameter should be added to modify this behavior, e.g `as_string`.

jmcarp: Sounds good. Maybe this isn't right, but I expect to be able to round-trip serializing and deserializing--it seems wrong if a serializer can't load what it dumps. Should `_deserialize` accept a list or a string for round-trip friendliness?

sloria: Yes, I think that's a fine solution.

jmcarp:
Updated.

webargs/fields.py
```diff
@@ -36,3 +36,33 @@ def __init__(self, nested, *args, **kwargs):
...
             nested = argmap2schema(nested)
         super(Nested, self).__init__(nested, *args, **kwargs)
+
+class DelimitedList(ma.fields.List):
+    """Same as `marshmallow.fields.List`, except can load from either a list or
+    a delimited string (e.g. "foo,bar,baz").
+
+    :param Field cls_or_instance: A field class or instance.
+    :param str delimiter: Delimiter between values.
+    :param bool dump_string: Dump values to string.
```
sloria: While I can see the reason for this naming, I think I'd rather this be `as_string` if only for consistency with `fields.Number`.

sloria:
Just one more suggested change.

webargs/fields.py
jmcarp: IMO `dump_string` is more explicit than `as_string`, but we can change this for consistency if you prefer.

jmcarp:
Updated--although for the record, I prefer the original parameter name :hamburger:

sloria:
I do think `dump_string` is slightly better, but I think consistency > correctness in this case. 

Anyway, thanks for this!

PR #75, opened 2015-10-23
Typo and dev environment step missing

imhoffd:
Two small things that I found while jumping around


sloria:
Test failure is unrelated to this change, and it is fixed on `dev`. Good to merge. 

Thanks!

PR #76, opened 2015-10-28
Ensure field data is available to marshmallow at the load_from location

DamianHeard:
webargs was storing data in the parsed dict at the destination location rather than at the load_from location as marshmallow expects.


DamianHeard:
Hmmm. My bad @sloria, I've stupidly bundled my bug fix and feature request into the same PR. No time right in this moment to sort this out. I will do in the coming days.

PR #77, opened 2015-11-05
use_args decorators now support callables in the first argument.

DamianHeard:
Any callable supplied should take the request as an argument and return a marshmallow schema instance.

### DIFF
--- a/webargs/pyramidparser.py
+++ b/webargs/pyramidparser.py
@@ -89,8 +89,9 @@ class PyramidParser(core.Parser):
         Supports the *Class-based View* pattern where `request` is saved as an instance
         attribute on a view class.
 
-        :param dict argmap: Either a `marshmallow.Schema` or a `dict`
-            of argname -> `marshmallow.fields.Field` pairs.
+        :param dict argmap: Either a `marshmallow.Schema`, a `dict`
+            of argname -> `marshmallow.fields.Field` pairs, or a callable
+            which accepts a request and returns a `marshmallow.Schema`.
         :param req: The request object to parse. Pulled off of the view by default.
         :param tuple locations: Where on the request to search for values.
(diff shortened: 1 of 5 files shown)


sloria:
Thanks for the PR! Looks like the build is failing due to a flake8 error. Can you please address this?

webargs/pyramidparser.py
```diff
@@ -112,8 +115,10 @@ def wrapper(obj, *args, **kwargs):
                     request = req or obj.request
                 except AttributeError:  # first arg is request
                     request = obj
-                parsed_args = self.parse(schema, req=request, locations=locations,
-                                         validate=validate, force_all=as_kwargs)
+
+                parsed_args = self.parse(schema or argmap(request), req=request,
```
sloria, line 119: I'm thinking this logic should go in `parse` rather than `use_args`. `parse`, `use_args`, and `use_kwargs` should all be able to take a schema factory as their first argument.

DamianHeard, line 119: I considered this but felt it wasn't overly useful since if I'm calling `parse` directly then I'm already going to have access to the request and therefore I can construct the schema precisely as I want. Is there some use case I'm missing or is it purely about pushing the logic in `parse` to save the duplication around the `schema or argmap(request)` style logic which presently occurs in a few places?

sloria, line 119: I think it makes sense to have consistent signatures among `parse`, `use_args`, and `use_kwargs`.

sloria, line 119: I am working on this now, so no further changes needed in this PR.

sloria:
Some modifications made during merge to allow `parse` to receive a factory.

Thanks again for the PR!

DamianHeard:
No thank you @sloria for producing such a great package.

PR #78, opened 2015-11-07
Add note in docs for nested fields and locations

imhoffd:
re: https://github.com/sloria/webargs/issues/74

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -198,39 +198,6 @@ To add your own custom location handler, write a function that receives a reques
     def posts(args):
         return 'displaying {} posts'.format(args['per_page'])
 
-.. _quickstart-custom-parsers:
-
-Custom Parsers
---------------
-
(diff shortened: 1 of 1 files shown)


sloria:
Don't worry about the test failure; it is unrelated to your changes.

sloria:
As noted in https://github.com/sloria/webargs/issues/74#issuecomment-154743501, `Nested` fields aren't  exclusive to JSON request bodies. Perhaps it would be useful to document an example of a custom parser that handles nesting in query params?

imhoffd:
Is this more of what you're thinking? The example I wrote up is a bit verbose, unfortunately.

sloria:
Looking good. Indeed, the example is a bit complicated for a quickstart, but it's a useful example nonetheless.

I just added a few "Advanced" topics to the quickstart. I think the "Custom Parser" section falls under that category. Can you update with the `dev` branch and put Custom Parser in the Advanced section?

Also, it might help to add a docstring to the custom parser, with an example of the expected URL format.

imhoffd:
Good call. How's it looking?

sloria:
Looks great. Thanks!

PR #79, opened 2015-11-08
slight tidy up to the pyramid parser to bring use_args into line

DamianHeard:
Functionally there's no change here. Just trying to make thing consistent.


sloria:
Good catch.

PR #83, opened 2015-12-21
Retain field name in error messages

immerrr:
This patch fixes a slight discrepancy between marshmallow and webargs error reporting: the former uses the original field name when reporting the error while the latter uses that of the field object.

In my specific case it meant that there was some effort required to rewrite, say, `content_type` returned in marshmallow's `ValidationError` to the original `Content-Type` to report the error to the user. I also had to add an option not to do that, as there was some pure-marshmallow parsing involved that had to be reported as well.

There's another discrepancy between marshmallow and webargs in interpreting `load_from`: marshmallow falls back to `load_from` only when the primary field name was not found, but something tells me that this was a conscientious decision (which I like more).

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -22,3 +22,4 @@ Contributors (chronological)
 * @marcellarius <https://github.com/marcellarius>
 * Damian Heard <https://github.com/DamianHeard>
 * Daniel Imhoff <https://github.com/dwieeb>
+* immerrr <https://github.com/immerrr>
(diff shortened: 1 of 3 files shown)


jacebrowning:
:+1:

I have also experienced this problem with `load_from` not being used in the error message.

sloria:
Hard to determine whether this is a breaking change or not. I think this is certainly the _desired_ behavior. I also think it is the _expected_ behavior, so I'm going to go ahead and include this in the next minor release.

Thanks for the PR!

Ross-Hunter:
This behavior is indeed better. However, this change did break stuff for me, as I was relying on the old behavior for error messages.

Edit: I just realized that this was a minor version bump, I should have locked myself down better.

PR #91, opened 2016-02-20
Closes #81: Added bulk-type arguments support

### LINKED ISSUE #81: Bulk-type arguments support
I'm not sure whether this case can be nicely handled with webargs, but anyway let me know if there is already an elegant solution to the issue I have.

I'm implementing a PATCH method in my RESTful API following [RFC 6902](http://tools.ietf.org/html/rfc6902), so the arguments would look like this:

```
PATCH /users/123

[
    { "op": "replace", "path": "/email", "value": "new.email@example.org" }
]
```

I have naively tried to pass `many=True` to my marshmallow schema, but it didn't work. Reading the code I couldn't find any hint that this might be a supported case.

Just for the sake of completeness, here is what was my naive attempt:

``` python
@use_args(MySchema(many=True))
def get(self):
    ...
```

I use Flask-RESTplus if this matters at all.

frol:
Read more details in #81.

Here is a brief example:

``` python
@use_args(MySchema(many=True), locations=('json', ))
def get(args):
```

Such function will expect to receive a JSON array of `MySchema` items.

### DIFF
--- a/webargs/aiohttpparser.py
+++ b/webargs/aiohttpparser.py
@@ -65,13 +65,12 @@ class AIOHTTPParser(AsyncParser):
     @asyncio.coroutine
     def parse_json(self, req, name, field):
         """Pull a json value from the request."""
-        if req.has_body and is_json_request(req):
-            json_data = self._cache.get('json')
-            if json_data is None:
-                self._cache['json'] = yield from req.json()
-            return core.get_value(self._cache['json'], name, field)
-        else:
+        if not (req.has_body and is_json_request(req)):
             return core.missing
+        json_data = self._cache.get('json')
+        if json_data is None:
+            self._cache['json'] = json_data = yield from req.json()
+        return core.get_value(json_data, name, field, allow_many_nested=True)
 
     def parse_headers(self, req, name, field):
         """Pull a value from the header data."""
--- a/tests/test_bottleparser.py
+++ b/tests/test_bottleparser.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 
 import pytest
-from bottle import Bottle, debug, request, response
+import json
+from bottle import Bottle, HTTPResponse, debug, request, response
 from webtest import TestApp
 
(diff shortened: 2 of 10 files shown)


webargs/aiohttpparser.py
```diff
@@ -68,10 +69,14 @@ def parse_json(self, req, name, field):
         if req.has_body and is_json_request(req):
             json_data = self._cache.get('json')
             if json_data is None:
-                self._cache['json'] = yield from req.json()
-            return core.get_value(self._cache['json'], name, field)
-        else:
-            return core.missing
+                self._cache['json'] = json_data = yield from req.json()
+            if json_data is not None:
+                if isinstance(json_data, dict):
```
sloria: Better to do `isinstance(json_data, collections.Mapping)` to handle all mapping types.

webargs/aiohttpparser.py
```diff
@@ -68,10 +69,14 @@ def parse_json(self, req, name, field):
...
+                    return core.get_value(json_data, name, field)
+                elif isinstance(json_data, list):
```
sloria: Better to handle any collection type here. You can use `marshmallow.utils.is_collection`.

sloria:
Thanks for the PR! The behavior that this implements looks correct. Is there a way we can get rid of the duplication across all the parsers?

frol:
@sloria Unfortunately, I don't have any sane idea of how to avoid the code duplication, but I will take a closer look tomorrow.

webargs/aiohttpparser.py
frol: OK, I will fix this.

frol:
@sloria I have moved the duplicated code into `core.get_value()` function introducing `allow_many_nested` flag for JSON parsers usage.

sloria:
Apologies for the delay in reviewing this again. I am currently on vacation; I will try to get this merged as soon as I am back.

frol:
Super! Thanks!

PR #95, opened 2016-03-22
Fix json parameter to support Flask 0.9

brettdh:
Flask 0.10 introduced the Request.get_json method, which webargs expects
to exist. This tiny change lets webargs work with Flask 0.9 to parse
JSON data in HTTP POST content.


sloria:
I am hesitant to promise that webargs will be completely compatible with Flask<=0.10.0 moving forward, but this change is AOK. Thanks!

PR #102, opened 2016-04-11
Docs about different ways of mixing locations

tuukkamustonen:


sloria:
Made some minor revisions in f2d31dc96. Thanks for the PR!

tuukkamustonen:
@sloria Huh, I was too hasty with that. Luckily you spotted the errors and fixed the grammary :+1:

PR #104, opened 2016-05-02
conda recipe for webargs

ijstokes:
With a conda recipe it will make it easier for Anaconda users to leverage webargs.  There is one challenge with this, which is exactly this issue: https://github.com/sloria/webargs/issues/86

But otherwise this provides a conda recipe that can build a conda package.


sloria:
I've decided against maintaining a conda recipe for now, as I'd like to keep the release process simple. Thanks for the PR, though!

PR #121, opened 2016-08-10
Fix flask.ext.restful deprecated warning

DStape:
"ExtDeprecationWarning: Importing flask.ext.restful is deprecated,
use flask_restful instead."

Api and Resource now imported from flask_restful.


DStape:
Probably jumped the gun on this one, I guess you'll want to support older versions of flask-restful.

sloria:
Thanks!

PR #123, opened 2016-09-22
Fix ValidationError causing HTTP/500 in Python 2.x

### LINKED ISSUE #122: ValidationError causes HTTP/500 in Flask instead of HTTP/422 in Python 2
Here are the symptoms: http://stackoverflow.com/questions/37321835/flask-error-handler-not-able-to-handle-assertion-error/39624893#39624893

Here is the case where I personally hit the issue with Python 2.x and PyPy in Python 2.x mode only: https://travis-ci.org/frol/flask-restplus-server-example/builds/161685419

After troubleshooting, I discovered that this issue stems from the `FlaskParser.handle_error` which calls `abort` function with `exc=` argument, which then [is set as "data"](https://github.com/sloria/webargs/blob/dev/webargs/flaskparser.py#L38) on the raised `HTTPException`, which in its turn is tried to be json-dumped and fails with `TypeError`, and now the new exception is in `sys.exc_info()`, which causes Flask to [assert here](https://github.com/pallets/flask/blob/01081dbe6cdfa3fc43d8e1fff708d4ed95e1be7e/flask/app.py#L1515).

Searching git history for the introduced `exc=` in `FlaskParser`, I found it here: https://github.com/sloria/webargs/commit/6f8088c7c85c05d2891e81ee97d38f2ae801159d

Is there a reason for `exc=` to be passed? How can we fix this?

Python 3 seems to have changed something with `sys.exc_info()` behaviour since even after the `TypeError` caused by `json.dumps()`, `sys.exc_info()` still reports `HTTPException` while in Python 2 it returns `TypeError`.

frol:
Closes #122.

### DIFF
--- a/tests/apps/flask_app.py
+++ b/tests/apps/flask_app.py
@@ -150,5 +150,4 @@ def echo_use_kwargs_missing(username, password):
 # Return validation errors as JSON
 @app.errorhandler(422)
 def handle_validation_error(err):
-    exc = err.data['exc']
-    return J({'errors': exc.messages}), 422
+    return J({'errors': err.exc.messages}), 422
--- a/tests/test_flaskparser.py
+++ b/tests/test_flaskparser.py
@@ -70,3 +70,19 @@ def test_abort_with_message():
     with pytest.raises(HTTPException) as excinfo:
         abort(400, message='custom error message')
     assert excinfo.value.data['message'] == 'custom error message'
+
+def test_abort_has_serializable_data():
+    with pytest.raises(HTTPException) as excinfo:
+        abort(400, message='custom error message')
+    serialized_error = json.dumps(excinfo.value.data)
+    error = json.loads(serialized_error)
+    assert isinstance(error, dict)
+    assert error['message'] == 'custom error message'
+
+    with pytest.raises(HTTPException) as excinfo:
+        abort(400, message='custom error message',
+            exc=ValidationError('custom error message'))
+    serialized_error = json.dumps(excinfo.value.data)
+    error = json.loads(serialized_error)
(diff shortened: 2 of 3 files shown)


frol:
OK, it seems that `exc` has some value, but maybe it can be put in a separate attribute?

frol:
Using this patched version of webargs my tests are back all green on Python 2.x.

sloria (changes requested):
I would like to see a regression test that reproduces the bug. I am not able to reproduce the error in the current test suite on Python 2 with Flask 0.11.1.

@frol Would you mind adding a regression test?

frol:
I'll try.

sloria:
Thanks so much!

frol:
OK, in fact, it seems to be Flask-RESTful/Flask-RESTplus specific as they handle the error in a way that just throw all `e.data` into JSON dump:
- Flask-RESTful - https://github.com/flask-restful/flask-restful/blob/63cadac8043e2070a22b5ab34a2656b804ef620f/flask_restful/__init__.py#L311
- Flask-RESTplus - https://github.com/noirbizarre/flask-restplus/blob/c0b25877b1089e354086a4a25f2f7e01f267baac/flask_restplus/api.py#L572

Do you still think that `data` should contain `exc`? I can go and create issues on those projects instead of "fixing" this here.

sloria:
Thanks for investigating @frol . I think your solution here is valid; `e.data` is meant to contain the data to be returned in JSON responses.

It would still be great to have a regression test for this.

frol:
I have added the test.

sloria (approved):

PR #144, opened 2016-11-15
original function in decorator

EFF:
Done:
===
Added a reference to the original function in the use_args decorator or unit testing purpose. Now, with these little changes, a user can import it's decorated function and still can unit test it's original function.

Should I had a small unit testing section in the docs ?

Note : 
---
I could have added the [Venusian](http://docs.pylonsproject.org/projects/venusian/en/latest/) library that is exactly for that purpose but it would have add a dependency to the project and it would have alter the decorators' behaviour a little. Tell me what you prefer.


sloria:
Thanks @EFF for the PR. Seems like a good change. I'll go ahead and merge this; feel free to send a follow-up PR with docs.

PR #148, opened 2017-01-08
Don't add missing fields as 'missing' to original_data

### LINKED ISSUE #146: Parser.parse adds "missing" to original_data
When parsing input, parser.parse adds `'field': missing` to `original_data` for every missing field. This happens only at first level, not in nested schema.

Hopefully, this test case will make this clearer:

```py
from marshmallow import Schema, fields, validates_schema

from webargs.flaskparser import parser

class Location(Schema):

    city = fields.Str()
    country = fields.Str()

class User(Schema):

    name = fields.Str()
    location = fields.Nested(Location)

    @validates_schema(pass_original=True)
    def print_original_data(self, data, original_data):
        print(original_data)

    class Meta:
        strict = True

input_1 = {'name': 'Eric Cartman'}
input_2 = {'location': {'city': 'South Park'}}

# Via pure Marshmallow
User().load(input_1)
User().load(input_2)
# {'name': 'Eric Cartman'}
# {'location': {'city': 'South Park'}}

# Mock Flask request object
class Request():
    def __init__(self, args):
        self.args = args

req_1 = Request(input_1)
req_2 = Request(input_2)

# Via webargs
parser.parse(User, req_1, locations=['query'])
parser.parse(User, req_2, locations=['query'])
# {'location': <marshmallow.missing>, 'name': 'Eric Cartman'}
# {'location': {'city': 'South Park'}, 'name': <marshmallow.missing>}
```

AFAIKT, this should not happen. I'd expect original_data to be the same as when using pure Marshmallow.

I could fix this by modifying `_parse_request`like this:

```py
            argdict = schema.fields
            parsed = {}
            for argname, field_obj in iteritems(argdict):
                parsed_value = self.parse_arg(argname, field_obj, req, locations)
                # If load_from is specified on the field, try to parse from that key
                if parsed_value is missing and field_obj.load_from:
                    parsed_value = self.parse_arg(field_obj.load_from, field_obj, req, locations)
                    argname = field_obj.load_from
                # Only add value if not missing
                if parsed_value is not missing:
                    parsed[argname] = parsed_value
```

Is this the correct thing to do?

sloria:
Fix #146

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -5,7 +5,7 @@ import sys
 
 import pytest
 import marshmallow
-from marshmallow import Schema, post_load, class_registry
+from marshmallow import Schema, post_load, class_registry, validates_schema
 from werkzeug.datastructures import MultiDict as WerkMultiDict
 
 PY26 = sys.version_info[0] == 2 and int(sys.version_info[1]) < 7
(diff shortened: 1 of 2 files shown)


sloria:
Build failure is unrelated to this change. I will fix the aiohttp errors in a separate PR.

PR #155, opened 2017-02-18
Update .gitignore

rgant:
Hide pytest failure .cache directory. 

Hide pyenv virtualenv file .python-version

This will just make my development life a little nicer as I try to contribute to webargs.


lafrech:
When I start a project, I generally get `.gitignore` from [here](https://github.com/github/gitignore/blob/master/Python.gitignore).

sloria:
Thanks!

PR #157, opened 2017-02-18
Add DelimitedPaths field

rgant:
This is a custom field especially meant to handle JSONAPI [include](http://jsonapi.org/format/#fetching-includes) and [sort](http://jsonapi.org/format/#fetching-sorting) parameters. I would have added it to marshmallow-jsonapi, but it is based on DelimtedList field which is only a part of webargs.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -857,6 +857,59 @@ def test_delimited_list_passed_invalid_type(web_request, parser):
         parser.parse(schema, web_request)
     assert excinfo.value.messages == {'ids': ['Not a valid list.']}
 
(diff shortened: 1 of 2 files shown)


sloria:
I'm lukewarm on this because of how specific to JSON-API to this API is. I understand why you added it to webargs, but I do think it belongs in marshmallow-jsonapi. Since delimited lists are common in the JSON-API spec, perhaps it would make sense to copy the `DelimitedList` field from webargs to `marshmallow_jsonapi.fields` and add `DelimitedPaths` there as well.

sloria:
Closing for now.

PR #158, opened 2017-02-23
Add pyramid's example

podhmo:
Hello. This is a example for pyramid.

### DIFF
--- a/examples/pyramid_example.py
+++ b/examples/pyramid_example.py
@@ -1,0 +1,79 @@
+"""A simple number and datetime addition JSON API.
+Run the app:
+
+    $ python examples/pyramid_example.py
+
+Try the following with httpie (a cURL-like utility, http://httpie.org):
+
+    $ pip install httpie
+    $ http GET :5001/
+    $ http GET :5001/ name==Ada
+    $ http POST :5001/add x=40 y=2
+    $ http POST :5001/dateadd value=1973-04-10 addend=63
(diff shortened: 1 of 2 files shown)


podhmo:
if TODO comments are a bit messy, I will remove it.

examples/pyramid_example.py
```diff
@@ -0,0 +1,83 @@
+"""A simple number and datetime addition JSON API.
+Run the app:
+
+    $ python examples/flask_example.py
```
sloria: This needs to be changed to `python examples/pyramid_example.py`

examples/pyramid_example.py
```diff
@@ -0,0 +1,83 @@
...
+from webargs import fields, validate
+from webargs.pyramidparser import use_args, use_kwargs
+
+
+hello_args = {
+    'name': fields.Str(missing='Friend')
+}
+
+
+# TODO: support
```
sloria: Can you remove these and report issues for the missing features you are referring to?

sloria (changes requested):
Looks good! Just a few tiny changes and this is good to merge.

examples/pyramid_example.py
podhmo: OK. I will fix.

podhmo:
fixed (and squashed).

sloria (approved):

sloria:
Thanks!

PR #163, opened 2017-04-27
Fix parse_form in AIOHTTPParser

DmitriyS:
aiohttp v2 dropped `POST` property from `BaseRequest`,
so req.POST raises exception


sloria:
Thanks!

PR #166, opened 2017-05-15
Fix for #165

### LINKED ISSUE #165: aiohttp parser breaks with more than one argument
Nothing seems to work once I add more than one field. Here's my minimal example.

I'm validating with:
```
programs_actions_args = {
    'type': webargs.fields.Str(required=True),
    'chain': webargs.fields.Integer(required=True),
}
```
and here's what happens when I POST:
```
POST /v1/programs/8137474d19926b0aa8efd4f1d3944131d59269d97a7bd8dab8e79d667eb314df/actions HTTP/1.1
Accept: application/json, */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 34
Content-Type: application/json
Host: localhost:8080
User-Agent: HTTPie/0.9.9

{
    "chain": "3",
    "type": "sampling"
}

HTTP/1.1 422 Unprocessable Entity
Content-Length: 46
Content-Type: application/json
Date: Fri, 05 May 2017 21:20:42 GMT
Server: Python/3.6 aiohttp/2.0.7

{
    "type": [
        "Missing data for required field."
    ]
}
```
Things work fine if there's only one argument.

zaro:
This PR fixes #165 for me.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -27,8 +27,9 @@ def test(ctx, coverage=False, browse=False):
 def flake(ctx):
     """Run flake8 on codebase."""
     cmd = 'flake8 .'
+    excludes = ['.direnv']
     if sys.version_info < (3, 4, 1):
-        excludes = [
+        excludes += [
             os.path.join('tests', 'apps', 'aiohttp_app.py'),
             os.path.join('tests', 'test_aiohttparser.py'),
             os.path.join('webargs', 'async.py'),
@@ -36,7 +37,7 @@ def flake(ctx):
             os.path.join('examples', 'annotations_example.py'),
             'build',
         ]
-        cmd += ' --exclude={0}'.format(','.join(excludes))
+    cmd += ' --exclude={0}'.format(','.join(excludes))
     ctx.run(cmd, echo=True)
 
 @task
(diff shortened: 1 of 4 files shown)


sloria:
Thanks @zaro . Would you mind adding a regression test for this? If you don't have time, let me know, and I can look into it over the next few days.

zaro:
@sloria I tried adding a regression test. But it doesn't seem to work properly.

It passes even if I revert the fix. Maybe you can give some clues how to make it fail when the fix is not applied.

I see that the test framework is actually faking an aiohttp server, so probably it has something to do with it.

I don't have much experience with aiohttp, but I imagine the problem is somehow connected to the fact that body in the request is usually streamed. The first time the check req.has_body succeeds and consumes the body , the following checks fail because the body is already consumed, and then this is not properly simulated in the webtest_aiohttp module.

tasks.py
```diff
@@ -27,16 +27,17 @@ def test(ctx, coverage=False, browse=False):
 def flake(ctx):
     """Run flake8 on codebase."""
     cmd = 'flake8 .'
+    excludes = ['.direnv']
```
sloria, line 30: This should go in `setup.cfg`.

sloria:
Ah, I think this bug will only appear in newer versions of aiohttp (>2.x). We are currently pinning to <2.0.0 in dev-requirements-py3.txt because webtest-aiohttp doesn't support 2.x yet. I've confirmed this fix by running `examples/aiohttp_example.py` after upgrading aiohttp, so I think this is good to merge. I'll fix the nitpick above after I merge.

PR #170, opened 2017-05-22
allow use_args/use_kwargs decorators with async def coroutines

zaro:
make the following work

``` python
@use_args({'name': fields.Str()})
async def api_test(request, args):
      ....
```

I am not certain this is the best way to achieve it,  but testing it  on my aiohttp application shows it works well.

### DIFF
--- a/webargs/async.py
+++ b/webargs/async.py
@@ -4,12 +4,21 @@ import asyncio
 import collections
 import inspect
 import functools
+import sys
 
 import marshmallow as ma
 from marshmallow.compat import iteritems
 from marshmallow.utils import missing
 
 from webargs import core
+
+PY_35 = sys.version_info >= (3, 5)
+
+
+def iscoroutinefunction(func):
+    if PY_35:
+        return inspect.iscoroutinefunction(func)
+    return False
 
 class AsyncParser(core.Parser):
     """Asynchronous variant of `webargs.core.Parser`, where parsing methods may be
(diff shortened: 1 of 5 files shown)


webargs/async.py
```diff
@@ -91,38 +95,55 @@ def use_args(self, argmap, req=None, locations=None, as_kwargs=False, validate=N
...
+                async def wrapper(*args, **kwargs):
+                    req_obj = req_
+
+                    # if as_kwargs is passed, must include all args
+                    force_all = as_kwargs
+
+                    if not req_obj:
+                        req_obj = self.get_request_from_view_args(func, args, kwargs)
+                    # NOTE: At this point, argmap may be a Schema, callable, or dict
+                    parsed_args = await self.parse(argmap,
```
sloria: In order for builds to pass on python < 3.5, I think you'll need to add `# flake8: noqa` to this line.

sloria (commented):
Thanks for the PR! I would definitely like to support `async def` coroutines. I am lukewarm on this because of the repeated code, but I can't think of a better way to do it at the moment.

Once the build is passing, this should be good to merge.

zaro:
I haven't had the time to properly check why it's failing on 3. 4. But I
would like to fix it and have this merged. Will do so in the next few days
I guess.

Regarding the code duplication I don't think there is other way for now.

sloria:
I believe it's failing flake8 on py34 `async`/`await` is only valid syntax in py>=3.5. As suggessted in https://github.com/sloria/webargs/pull/170#discussion_r119502721 , you can just tell flake8 to ignore that line.

zaro:
Yeah, but if it's failing flake8 with 3.4 it will fail real usage with
python3.4 also.

I was thinking to split it the decorator in separate file, having two
variants one for 3.4 and one for 3.5+ and import based on currently running
python version.

sloria:
Sounds like a good approach.

zaro:
Seems I managed to find the right flake8 exclude list for all tests to pass :)

sloria:
Excellent. Thanks @zaro! I've added tests for this in #172 . Once those pass, this will be good to merge.

PR #179, opened 2017-07-18
Bug fix: Added support for using schema as kwargs.

Itayazolay:
Fixed bug: set `as_kwargs=True` when using Schema raises an AttributeError.


sloria:
Thanks!

Issue #173, opened 2017-06-20
Parser passes incomplete data to marshmallow.pre_load schema methods


benkogan:
When using a Marshmallow schema with a `@pre_load`-decorated method, the `data` passed as the arg to the method includes only fields specified in the schema. This breaks pre_load methods that depend on data entires not in fields such as [the enveloping example given in the Marshmallow docs][env].

Here's a minimal example using Flask to demonstrate the issue:

```python
from flask import Flask, jsonify
from marshmallow import Schema, fields, pre_load
from webargs.flaskparser import use_args

class Args(Schema):
    foo = fields.String()

    class Meta:
        strict = True

    @pre_load
    def unwrap(self, data):
        return data['data']

app = Flask(__name__)

@app.route('/', methods=['POST'])
@use_args(Args)
def index(args):
    print('got: {}'.format(args))
    return jsonify(dict(got=args))

if __name__ == '__main__':
    app.run(port=5001, debug=True)
```

When sending data to the route above (for example using httpie: `http POST :5001/ data:='{"foo": "bar"}'` the following trace is produced:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 2000, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1991, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1567, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.4/dist-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.4/dist-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
[...]

PR #201, opened 2018-02-20
Allow Marshmallow validation

klapaukh:
This PR attempts to address problems raised in #173 that Marshmallow validation methods never get to see the full set of arguments passed, even if pass_original is given. 
This PR changes the way that webargs reads arguments from the request, changing argument retrieval to create a full dictionary of possible parameters

This does introduce a problem with the webapp2 parser where the headers are case-insensitive, but that is probably addressable if the general idea of the PR is okay.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ pip-log.txt
 .tox
 nosetests.xml
 htmlcov
+.pytest_cache/
 
 # Translations
 *.mo
@@ -42,6 +43,9 @@ output/*/index.html
 
 # Sphinx
 docs/_build
+
+# VSCode 
+.vscode/
 
 README.html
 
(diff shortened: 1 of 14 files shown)


.gitignore
```diff
@@ -43,6 +44,9 @@ output/*/index.html
 # Sphinx
 docs/_build
 
+# VSCode
```
lafrech: Don't adapt a project .gitignore to your environment.

You should use a global .gitignore for that: https://help.github.com/articles/ignoring-files/

klapaukh: Oh, sorry, I wasn't paying attention. Now restored.

sloria:
Closing, as this PR has gone quite stale.

We do still want to implement this feature, however, so feel free to rework this and open a new PR.

PR #211, opened 2018-03-30
In marshmallow 3, data_key replaces load_from

lafrech:
It could look like there is duplication in core. This is because the change is not only the parameter renaming, but also the fact that in MA3, the attribute name is not checked when `data_key` is specified, which allows to simplify the logic.

Besides, in those cases, I prefer duplication to complicated logic because ultimately, the old compatibility code will be dropped and we'll be left with the simple case.


sloria:
Looks good! Thanks.

PR #222, opened 2018-04-30
Minor refactor in core.py

### LINKED ISSUE #216: Dead code in core.py?
I just noticed what seems to be dead code.

1/ Unless I'm mistaken, the `instance` and `kwargs` attributes are both untested and unused in the code.

https://github.com/sloria/webargs/blob/dev/webargs/core.py#L87

```python
    def argmap2schema(argmap, instance=False, **kwargs):
```

2/ In `load`, `argmap` is always a schema instance due to the call to `_get_schema` in `parse`, so the test is useless.

```python
    def load(self, data, argmap):
        if isinstance(argmap, ma.Schema):
            schema = argmap
        else:
            schema = argmap2schema(argmap)()
```

In both cases, all tests pass when removing suspected dead code.

lafrech:
Closes #216.

This simplifies the code a bit.

No test broken. However, any custom parser that overrides `Parser.parse` (`or AsyncParser.parse`) and calls `Parser.load` requires a little adaptation.

Does that make it a breaking change? The docs read

> To add your own parser, extend Parser and implement the parse_* method(s) you need to override. 

`Parser.parse` and `Parser.load` are not exposed, so it can be argued that the API is not changed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+3.0.0 (unreleased)
+******************
+
+Changes :
+
+* Remove unused ``instance`` and ``kwargs`` arguments of ``argmap2schema``
+* Remove ``Parser.load`` method (``Parser`` now calls ``Schema.load`` directly)
+
+Those changes shouldn't affect most users. However, they might break custom parsers calling these methods. (:issue: `222`)
+
 2.1.0 (2018-04-01)
 ******************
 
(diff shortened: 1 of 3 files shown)


sloria:
I'm fine with this change. I think it's a breaking change, though, since we're removing a public method of a public class. We should try our best to follow semver as conservatively as is practical. We can make clear in the CHANGELOG that this change shouldn't affect most users.

lafrech:
I just updated the changelog.

sloria:
Thanks. The failing build is unrelated to this change, so this is good to merge.

PR #232, opened 2018-06-06
Respect DEFAULT_VALIDATION_STATUS set on Parser

sloria:
see https://github.com/sloria/webargs/issues/180#issuecomment-394869645

### DIFF
--- a/webargs/core.py
+++ b/webargs/core.py
@@ -183,7 +183,9 @@ class Parser(object):
     """
 
     DEFAULT_LOCATIONS = ('querystring', 'form', 'json',)
+    #: Default status code to return for validation errors
     DEFAULT_VALIDATION_STATUS = DEFAULT_VALIDATION_STATUS
+    #: Default error message for validation errors
     DEFAULT_VALIDATION_MESSAGE = 'Invalid value.'
 
     #: Maps location => method name
--- a/webargs/core.py
+++ b/webargs/core.py
@@ -297,12 +297,15 @@ class Parser(object):
         if (isinstance(error, ma.exceptions.ValidationError) and not
                 isinstance(error, ValidationError)):
             # Raise a webargs error instead
+            error_kwargs = getattr(error, 'kwargs', {})
+            if 'status_code' not in error_kwargs:
+                error_kwargs['status_code'] = self.DEFAULT_VALIDATION_STATUS
             error = ValidationError(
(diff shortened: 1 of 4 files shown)


webargs/core.py
```diff
@@ -297,12 +299,15 @@ def _on_validation_error(self, error, req):
         if (isinstance(error, ma.exceptions.ValidationError) and not
                 isinstance(error, ValidationError)):
             # Raise a webargs error instead
+            error_kwargs = getattr(error, 'kwargs', {})
+            if 'status_code' not in error_kwargs:
```
lafrech, line 303: Can you provide a use case example where `status_code` is in kwargs?

Actually, I don't see the use case for attaching `kwargs` to the `ValidationError`.

sloria: The `kwargs` passed to `ValidationError` need to be attached because ValidationErrors get *re-raised* after schema validation is complete.

Take this route in the [tests](https://github.com/sloria/webargs/blob/616d02acefe40c465b14e792322483a7546d5055/tests/apps/flask_app.py#L78-L83):

```python
@app.route('/error400', methods=['GET', 'POST'])
def error400():
    def always_fail(value):
        raise ValidationError('something went wrong', status_code=400)
    args = {'text': fields.Str(validate=always_fail)}
    return J(parser.parse(args))
```

webargs creates a marshmallow `Schema` from the `args` dictionary then validates the request against it. When the `ValidationError` gets raised in the validator, marshmallow collects the error, with its kwargs. The `marshmallow.ValidationError` that gets raised gets caught by webargs and [reraised as a webargs.ValidationError](https://github.com/sloria/webargs/blob/616d02acefe40c465b14e792322483a7546d5055/webargs/core.py#L299-L300).

lafrech: OK, thanks. I think I get the mechanism. I didn't see a use case because my typical and only use case is to define `Schema`s to [de|]serialize my model and although those `Schema`s belong to the API part of the app, they don't play with `status_code` and other web related stuff, I leave this to webargs, so they just raise raw `ValidationError`s. Or sometimes in my views, I'll catch an error and call `abort` passing a specific status code explicitly.

Anyway, thanks for the example. Fair enough.

lafrech:
LGTM.

Just a typo in the CHANGELOG: Suppor -> Support.

PR #240, opened 2018-07-07
Python 3.7 async and await are now reserved keywords.

Reskov:
At python 3.7 import webargs async module will raise following syntax exception

```
from webargs.async import AsyncParser
  File "<input>", line 1
    from webargs.async import AsyncParser
                     ^
SyntaxError: invalid syntax
```


sloria:
Thanks for reporting @Reskov . Looks like we need to rename that module. We'll have this fixed sometime this weekend.

sloria:
If you'd like to send a PR renaming `webargs.async` -> `webargs.asyncparser`, be my guest! If you don't have time, I can take care of it this weekend.

sloria:
Thanks @Reskov !

PR #244, opened 2018-07-09
Format code and docs with black+blacken-docs+pre-commit

sloria:
* Format code with [black](https://github.com/ambv/black)
* Format doc code blocks with [blacken-docs](https://github.com/asottile/blacken-docs/). This uncovered a number of syntax issues in the docs.
* Run both with [pre-commit](https://pre-commit.com/)


sloria:
@asottile If you're feeling generous, feel free to give this a drive-by =).

docs/conf.py
```diff
@@ -6,51 +6,51 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('..'))
-import webargs
+sys.path.insert(0, os.path.abspath(".."))
+import webargs  # flake8: noqa
```
asottile: [fun story about `# flake8: noqa`](https://gitlab.com/pycqa/flake8/merge_requests/219)

I _believe_ you want `# noqa` here (or `# flake8: noqa` on a line by itself if you actually don't want to lint this file)

setup.cfg
```diff
@@ -2,6 +2,8 @@
 universal = 1
 
 [flake8]
-ignore = E127,E128,E265,E302,N803,N804,N806,E731,E402,E266,E305
-max-line-length = 100
-exclude = .git,.ropeproject,.tox,build,.direnv,__pycache__
+ignore = E203, E266, E501, W503, E302
+max-line-length = 80
```
asottile: black I believe uses `88`, though I see you've ignored `E501` so I guess this line does nothing! :D

tasks.py
```diff
@@ -5,79 +5,95 @@
...
     retcode = pytest.main(args)
     if coverage and browse:
-        webbrowser.open_new_tab(os.path.join('htmlcov', 'index.html'))
+        webbrowser.open_new_tab(os.path.join("htmlcov", "index.html"))
     sys.exit(retcode)
 
+
+@task
+def precommit(ctx):
+    ctx.run("pre-commit run --all-files", echo=True)
```
asottile: `--show-diff-on-failure` is usually a good option for CI's sake (helps drive-by contributors that might not have set up the linting locally)

asottile (commented):
Looks good!  Just a few small things I noticed :)

setup.cfg
lafrech: Is it possible to use explicit names in there?

    -ignore = E203, E266, E501, W503, E302
    +ignore = rule-number-one, another-rule

lafrech:
> Black can be installed by running pip install black. It requires Python 3.6.0+ to run but you can reformat Python 2 code with it, too.

My local environment is Python 3.5. Will this make `pre-commit` fail?

setup.cfg
asottile: unfortunately not -- pylint has this but flake8 does not

asottile:
> My local environment is Python 3.5. Will this make `pre-commit` fail?

asottile:
> My local environment is Python 3.5. Will this make `pre-commit` fail?

Not necessarily, the hooks here are configured with `language_version: python3.6` so pre-commit will attempt to instantiate hook environments using that interpreter -- if there's no python3.6 then yes it will fail hook installation and tell you so.

setup.cfg
sloria: > black I believe uses 88, though I see you've ignored E501 so I guess this line does nothing! :D

This is from the [recommended flake8 config from black](https://github.com/ambv/black/blob/master/.flake8). E501 is ignored, but Bugbear's B901 is enabled, which will warn when the max-line-length exceeds 10% of `max-line-length`.

docs/conf.py
sloria: Ah, TIL. Thanks.

sloria:
Addressed @asottile 's comments and also added some notes in CONTRIBUTING.rst about needing the `python3.6` interpreter for `pre-commit` to work properly. If you don't have it installed, you can still run tests with 

```
inv test --no-syntax
```

lafrech:
OK. This means that all contributors on Python 3.5 won't be able to use `pre-commit` anymore.

There's a trade-off, here. This commit adds more features to `pre-commit` but prevents part of the contributors to use it at all. I don't have stats on who uses 3.5 or 3.6. Using Debian Stable, I'll probably be using 3.5 for a while.

No big deal. If I can't run it locally, Travis will complain and I'll use my special ability (`git push -f`) to fix my mess.

sloria:
@lafrech You can use Python 3.5 in your virtualenv and still use pre-commit as long as you have the `python3.6` interpreter globally.

With pyenv:

```
pyenv install 3.6.5
pyenv global system 3.6.5
```

setup.cfg
asottile: are you using `flake8-bugbear`? I don't see it installed

sloria: Ah, woops. Forgot to add that to dev-requirements.txt. Will fix this up after work today.

PR #297, opened 2018-09-25
aiohttpparser: Fix 500 error with JSON content-type and empty body

fmqa:
Fixes 500 errors when Content-Type is JSON and the request body is empty.
This is done by catching the resulting JSONDecodeError and checking whether the input is empty.

### DIFF
--- a/tests/test_aiohttpparser.py
+++ b/tests/test_aiohttpparser.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 
 import asyncio
+import webtest
 import webtest_aiohttp
 import pytest
 
+from io import BytesIO
 from webargs.core import MARSHMALLOW_VERSION_INFO
 from webargs.testing import CommonTestCase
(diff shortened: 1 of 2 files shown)


sloria:
Sorry for the delay in getting to this. I wasn't sure if this was correct given that aiohttp decided not to fix the issue on their end: https://github.com/aio-libs/aiohttp/issues/3302 .

But I think it is correct to handle an empty body in webargs since it has its own `missing` value and the decorator API.

@kochab Would you mind adding a test for this? If you don't have time, let me know, and I can do it when I have some free time.

Once the test is added, this is good to merge.

fmqa:
Hi @sloria. No problem! I have added a test for this as requested. It doesn't look idiomatic, but I did it this way because `webtest.TestApp.post()` adds an unwanted `Content-Length` header. `webtest.TestRequest.blank()` can be used to build a minimal request with just a `Content-Type` header and an empty/zero-length body.

PR #299, opened 2018-10-04
Bump marshmallow requirement version to >=2.15.2

lafrech:
This version includes fixes for

- https://github.com/marshmallow-code/marshmallow/issues/772 (Empty Only Treated as None) [CVE-2018-17175]
- https://github.com/marshmallow-code/marshmallow/issues/785 (Thread-safety issue in error structure)

Hopefully, this should make Tidelift happy.

While upgrading to 2.15.1 for CVE-2018-17175, I figured I'd go to 2.15.2 as the thread safety issue in marshmallow can be problematic when used with webargs to parse inputs in a threaded server (this is how we discovered the problem).

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 # Requirements
-marshmallow
+marshmallow>=2.15.2
 
 # Task running
 invoke==1.2.0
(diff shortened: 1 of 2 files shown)


sloria:
Good call. I was wondering how to address the Tidelift dependency build error. Hope this does the trick.

Issue #252, opened 2018-07-18
Deleted fields are always recreated with use_kwargs


piroux:
If one tries to `load` a json object with Marshmallow so that it validates a field but one does not wish to have it in the final loaded object, one can have it removed using a method decorated with `post_load` so that the field is validated but not exported:

```python
     class GeojsonFeatureCollection(Schema):
         geotype = fields.String(required=True, location='json', load_from='type',
                                 validate=validate.OneOf(choices=['FeatureCollection']))
         features = fields.List(fields.Dict())

         @post_load
         def remove_geotype(self, data):
             data.pop('geotype')
             return data

         class Meta:
             strict = True
```

It works well with Marshmallow.

However, it is not working with webargs. For instance here, it will recreate the geotype member in the dictionary with  the `<marshmallow.missing>` value. Especially, the use of the `use_kwargs` decorator in the view method will force webargs to recreate all the missing fields as one can see here in the parser with the call to the `fill_in_missing_args` function: https://github.com/sloria/webargs/blob/a8640995dbe423d30248a26474f9fb12bad997f6/webargs/core.py#L399-L401

Indeed, the `force_all` argument of the `parse` method is currently always equal to `True` when one uses the `use_kwargs` decorator: https://github.com/sloria/webargs/blob/a8640995dbe423d30248a26474f9fb12bad997f6/webargs/core.py#L464-L475

I think one shoud be able retain in webargs the behaviour that Marshmallow provides so that the deleted fields would not be recreated by default if a flag is passed.

A simple solution would be to export the `force_all` variable as a parameter, maybe with a more explicit name, like `force_recreate_missing_fields` to the `use_args` method so that  one could use it from `use_kwargs`.

Moreover, I am wondering why was it required to recreate all the missing fields if the arguments have to be inserted as keyword arguments ? Was it to avoid raising an error in the views methods in case one argument was missing ? 
If one adds this parameter offering more control over the missing fields, I think it would make sense now not to recreate the missing fields by default, right ? (have `force_recreate_missing_fields=False in `use_args` as parameter)

sloria:
I've added the `force_all` argument to `use_kwargs`. It will be released in 4.2.0.

I suggest we remove the `fill_in_missing_args` behavior altogether. I've opened an RFC: #342.

PR #307, opened 2018-10-22
fix use_kwargs used with ma.Schema(partial=True|tuple)

TTWShell:
We hope ignore missing fields when instance `ma.Schema` with `partial=True`.  
See https://marshmallow.readthedocs.io/en/3.0/api_reference.html#schema 
#308

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -853,43 +853,36 @@ def test_use_kwargs_with_arg_missing(web_request, parser):
     assert viewfunc() == {"username": "foo", "password": missing}
 
 
-def test_use_kwargs_with_arg_missing_and_partial(web_request, parser):
+@pytest.mark.parametrize(
+    "partial, expected_result",
+    [
+        (True, {"email": "foo@bar.com"}),
+        (["email"], {"email": "foo@bar.com", "password": missing}),
+        (["password"], {"email": "foo@bar.com"}),
+    ],
+)
+def test_use_kwargs_with_arg_missing_and_partial(
+    web_request, parser, partial, expected_result
+):
     class UserSchema(Schema):
         id = fields.Int(dump_only=True)
         email = fields.Email()
         password = fields.Str(load_only=True)
+
         if MARSHMALLOW_VERSION_INFO[0] < 3:
 
             class Meta:
                 strict = True
 
-    # test partial is True
     web_request.json = {"email": "foo@bar.com"}
     kwargs = deepcopy(strict_kwargs)
-    kwargs["partial"] = True
+    kwargs["partial"] = partial
 
     @parser.use_kwargs(UserSchema(**kwargs), web_request)
     def viewfunc(**kwargs):
         return kwargs
 
-    assert viewfunc() == {"email": "foo@bar.com"}
-
-    # test partial is list or tuple
-    kwargs["partial"] = ["email"]
-
-    @parser.use_kwargs(UserSchema(**kwargs), web_request)
-    def viewfunc(**kwargs):
-        return kwargs
-
-    assert viewfunc() == {"email": "foo@bar.com", "password": missing}
-
-    kwargs["partial"] = ["password"]
-
-    @parser.use_kwargs(UserSchema(**kwargs), web_request)
-    def viewfunc(**kwargs):
-        return kwargs
-
-    assert viewfunc() == {"email": "foo@bar.com"}
+    assert viewfunc() == expected_result
 
 
 def test_delimited_list_default_delimiter(web_request, parser):
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from copy import deepcopy
 import itertools
 import mock
 import sys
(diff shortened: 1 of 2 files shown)


lafrech:
Looks good to me.

You could parametrize the test to reduce duplication (pass (partial, expected_result) as tuple).

TTWShell:
@lafrech Thank you for taking your time for the review. Did you mean this?

	def test_use_kwargs_with_arg_missing_and_partial(web_request, parser):
	    class UserSchema(Schema):
	        id = fields.Int(dump_only=True)
	        email = fields.Email()
	        password = fields.Str(load_only=True)
	        if MARSHMALLOW_VERSION_INFO[0] < 3:

	            class Meta:
	                strict = True

	    web_request.json = {"email": "foo@bar.com"}
	    kwargs = deepcopy(strict_kwargs)

	    def assert_excepted(partial, expected_result):
	        kwargs['partial'] = partial

	        @parser.use_kwargs(UserSchema(**kwargs), web_request)
	        def viewfunc(**kwargs):
	            return kwargs

	        assert viewfunc() == expected_result

	    assert_excepted(True, {"email": "foo@bar.com"})
	    assert_excepted(["email"], {"email": "foo@bar.com", "password": missing})
	    assert_excepted(["password"], {"email": "foo@bar.com"})

lafrech:
I meant something like that. I could have been more explicit.

parametrize is probably my favorite pytest feature.

```py
@pytest.mark.parametrize('case', [
    (True, {"email": "foo@bar.com"})
    (["email"], {"email": "foo@bar.com", "password": missing}),
    (["email"], {"email": "foo@bar.com"}),
])
def test_use_kwargs_with_arg_missing_and_partial(web_request, parser, case):
    partial, expected = params

    class UserSchema(Schema):
        id = fields.Int(dump_only=True)
        email = fields.Email()
        password = fields.Str(load_only=True)
        if MARSHMALLOW_VERSION_INFO[0] < 3:
            class Meta:
                strict = True

    web_request.json = {"email": "foo@bar.com"}
    kwargs = deepcopy(strict_kwargs)
    kwargs["partial"] = partial

    @parser.use_kwargs(UserSchema(**kwargs), web_request)
    def viewfunc(**kwargs):
        return kwargs
    assert viewfunc() == expected

```

TTWShell:
@lafrech updated and test passed.

tests/test_core.py
```diff
@@ -852,6 +853,38 @@ def viewfunc(username, password):
...
+        if MARSHMALLOW_VERSION_INFO[0] < 3:
+
+            class Meta:
+                strict = True
+
+    web_request.json = {"email": "foo@bar.com"}
+    kwargs = deepcopy(strict_kwargs)
+    kwargs["partial"] = partial
+
+    @parser.use_kwargs(UserSchema(**kwargs), web_request)
```
lafrech: Suggestion:

- remove `import deepcopy`

- `UserSchema(partial=partial, **kwargs)`

sloria (changes requested):
I'm unsure about this change.

With this change, the number of arguments to the decorated view will vary. Therefore you can't use named arguments.

```python
# won't work
@parser.use_kwargs(UserSchema(partial=True))
def viewfunc(request, username, password):
```

You could, of course, use `**kwargs` instead of named arguments, but at that point you might as well use `use_args` instead.

Am I missing something?

TTWShell:
Thanks.

We want used like this:

	class UserSchema(ModelSchema):
		username =  fields.Str(required=True)
		password =  fields.Str(required=True)
		# other optinal fields auto generated by ModelSchema ...


	@bp.route('/users', methods=['POST'])
	@use_kwargs(UserSchema(partial=True))
	create_user(username, password, **kwargs):
	    User(username=username, password=password, **kwargs)


Named arguments mean that required, others in kwargs as optinal.

Explicit is better than implicit.

lafrech:
You could do 

```py
class UserSchema(ModelSchema):
	username =  fields.Str(missing=username)
	password =  fields.Str(missing=password)
	# other optional fields auto generated by ModelSchema ...

@bp.route('/users', methods=['POST'])
@use_kwargs(UserSchema(partial=True))
create_user(**kwargs):
    User(**kwargs)
```

I didn't really think about the use case. I don't think this change forces anything on users (if you don't use partial, it makes not difference, and if you do, the new behaviour is more consistent).

AFAIU, the question here is whether using partial is a good idea.

TTWShell:
partial (bool|tuple) – Whether to ignore missing fields. If its value is an iterable, only missing fields listed in that iterable will be ignored.

`use_kwargs` force fill_in_missing_args, so `partial` won't work as excepted.

    @use_kwargs(UserSchema(partial=False))
    create_user(username, password, **kwargs):
        User(username=username, password=password, **kwargs)

`kwargs` will be contain `marshmallow.utils.missing` value field. This is not expected.

GreenSmile:
Guys, there is similar issue related to `force_all` and `fill_in_missing_args()`: https://github.com/sloria/webargs/issues/252

I believe the proper solution for both issues would be extra `use_kwargs()` parameter to ignore `force_all`. This way you will be able to disable `fill_in_missing_args()`, so any changes applied to data by Marshmallow will be preserved. This  solution is backward compatible and solves both issues at once. Something like:

```
@use_kwargs(UserSchema(), partial=True)  # Note: partial is parameter of use_kwargs.
def create_user(username, password, **kwargs):
    User(username=username, password=password, **kwargs)
```

or as I use it:
```
@use_kwargs(UserSchema(), partial=True)
def update_user(user_id, **kwargs):
    user = read_user(user_id)
    for key, value in kwargs.items():
        setattr(user, key, value)
    user.save()
```

sloria:
I think I'm leaning towards @GreenSmile 's suggestion. Exposing a way to disable the `fill_in_missing_args()` call is a minimally impactful change, and it solves the OP and #252. I'm not sure about the name `partial` for the new param, though--users may confuse it with marshmallow's `partial`. ~~Maybe `fill_missing`?~~ `Parser.parse` already has a `force_all` param; how about we expose it to use_args/use_kwargs?

@TTWShell @lafrech Thoughts?

EDIT: Suggest `force_all` instead of `fill_missing`.

sloria:
I've added the `force_all` argument to use_kwargs to provide a backwards-compatible solution to this issue and #252. So I'm going to close this for now.

I think we should just remove the `fill_in_missing_args` behavior. I've opened an RFC: #342. Please comment there if you have thoughts/opinions.

PR #319, opened 2018-11-01
fix DelimitedList serialize bug

lee3164:
DelimitedList should use it's parents serialize results when as_string is set
Change-Id: Ic7318703fce93c64373eabf1737e2480256eb557

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -882,18 +882,25 @@ def test_delimited_list_as_string(web_request, parser):
 
 
 def test_delimited_list_as_string_v2(web_request, parser):
-	web_request.json = {"dates": "2018-11-01,2018-11-02"}
-	schema_cls = argmap2schema(
-		{"dates": fields.DelimitedList(fields.DateTime(format="%Y-%m-%d"), as_string=True)}
-	)
-	schema = schema_cls()
+    web_request.json = {"dates": "2018-11-01,2018-11-02"}
+    schema_cls = argmap2schema(
+        {
+            "dates": fields.DelimitedList(
+                fields.DateTime(format="%Y-%m-%d"), as_string=True
+            )
+        }
+    )
+    schema = schema_cls()
 
-	parsed = parser.parse(schema, web_request)
-	assert parsed["dates"] == [datetime.datetime(2018, 11, 1), datetime.datetime(2018, 11, 2)]
+    parsed = parser.parse(schema, web_request)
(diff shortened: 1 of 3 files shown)


lafrech:
Thanks.

Do you think you could add a test, or rework `test_delimited_list_as_string` to actually test your fix?

lee3164:
ok, I add a new test case and test is pass

tests/test_core.py
```diff
@@ -880,6 +881,22 @@ def test_delimited_list_as_string(web_request, parser):
     assert data["ids"] == "1,2,3"
 
 
+def test_delimited_list_as_string_v2(web_request, parser):
+	# b26793ee8021b6cfe525ac4f05570c3c68a4debd commit info
```
sloria: Nitpick: No need to include the commit info here.

sloria (changes requested):
Looks like `pre-commit` step failed. Running `invoke precommit` should fix the issue. Feel free to amend your latest commit with the formatting change.

sloria:
Also, please add yourself to AUTHORS.rst.

tests/test_core.py
lee3164: 🤣，I just want to others know why I add this test case

sloria: That's what `git blame` is for =)

lee3164: ok, i will delete this line

lee3164: why CI cannot pass?

sloria: See my comment above. You need to format the code using `invoke precommit`.

sloria: Let me know if you have trouble getting the command to work. I can help you work through the errors if you want.

Or, if you don't have time, I'd be glad to make the change myself.

lee3164: I very appreciate your kindness. I do have some troubles in this command. I do not know how to execute this command. So, could you tell the steps to work it?

sloria: In a Python 3.6 virtual environment, install the dev-requirements:

```
pip install -r dev-requirements.txt
```

Then run

```
invoke precommit
```

This will format the code using `black`. Then commit and push the changed files.

lee3164: Thank you very much, I never make a pr before

sloria: That's great! Thank *you* for taking the time to make the PR.

sloria (approved):

PR #326, opened 2018-11-22
Fixed bug with using multiple use_args decorators on the same aiohttp handler.

jangelo:
This PR is to address the following issue:
https://github.com/sloria/webargs/issues/273

The issue with the previous code was when multiple use_args are present, the function `get_request_from_view_args` would be called for each one.  In this function, when the len(args) > 1, it would just take the value at arg[1] and assume it was the web.Request object.  This unfortunately is not the case when multiple use_args exist.  Instead, the fix is to scan the args list and find the web.Request object.

A unittest was added copying the use case as specified in the documentation.
https://webargs.readthedocs.io/en/latest/advanced.html#mixing-locations


sloria (approved):
Looks good. Thanks!

PR #328, opened 2018-11-22
Remove webargs.ValidationError and use marshmallow ValidationError instead

### LINKED ISSUE #327: RFC: Don't pass a status_code as kwargs in a Field validator
webargs's `ValidationError` adds `status_code` and `headers` to marshmallow's `ValidationError`. This allows to specify a default code for input validation error. The webserver uses this attribute to return the proper error code.

The tests show another use of this that I find a bit problematic.

It is possible to define a validator on a `Field` of the input validation `Schema` that raises a `ValidationError` with a `status_code`. The deserialization logic in marshmallow adds this attribute to the `ValidationError` that is raised at the end of the validation process.

The issue with this is the case of several fields raising with a different status code. The last one to deserialize has priority. This is probably a corner case, but it shows the feature is not 100% reliable. It looks underterministic.

In practice I don't see the use case for this in the first place. I generally use the default 422 for all validation errors and return specific codes from flask abort in view function or in specific decorator (authentication decorator,...).

Related marshmallow issue: https://github.com/marshmallow-code/marshmallow/issues/996.

My suggestion is to remove the logic that passes the kwargs in marshmallow and delete the test case here that exhibit the case of specifying a status code in a field validator.

lafrech:
Closes #327.

With this change, the naming of `DEFAULT_VALIDATION_STATUS` is a bit improper, but we can leave it as is to avoid breaking stuff everywhere.

### DIFF
--- a/webargs/core.py
+++ b/webargs/core.py
@@ -14,6 +14,7 @@ except ImportError:
     import json
 
 import marshmallow as ma
+from marshmallow import ValidationError
 from marshmallow.compat import iteritems
 from marshmallow.utils import missing, is_collection
 
@@ -21,7 +22,6 @@ logger = logging.getLogger(__name__)
 
 
 __all__ = [
-    "WebargsError",
     "ValidationError",
     "argmap2schema",
     "is_multiple",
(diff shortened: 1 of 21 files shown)


webargs/core.py
```diff
@@ -54,9 +54,7 @@ def __init__(
     ):
         self.status_code = status_code
```
sloria: This should be removed, since it will no longer be respected.

sloria (changes requested):
Just one suggested change.

There are a few tests in `tests/test_core.py:TestValidationError` will need to be modified/removed.

lafrech:
Just rebased.

I also went ahead with the removal of `webargs.ValidationError`. It results in further code simplification (I mean removal) in `_on_validation_error`. Nice.

sloria (approved):

PR #330, opened 2018-11-30
Fix marshmallow 3.0.0rc1 compatibility

lafrech:
See https://github.com/marshmallow-code/marshmallow/issues/840#issuecomment-443142531.

### DIFF
--- a/webargs/core.py
+++ b/webargs/core.py
@@ -6,6 +6,7 @@ import functools
 import inspect
 import logging
 import warnings
+from distutils.version import LooseVersion
 
 try:
     import simplejson as json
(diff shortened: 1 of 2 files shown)


webargs/core.py
```diff
@@ -323,10 +323,12 @@ def _on_validation_error(self, error, req, schema):
         ):
             # Raise a webargs error instead
             kwargs = getattr(error, "kwargs", {})
-            kwargs["field_names"] = error.field_names
             kwargs["data"] = error.data
             if MARSHMALLOW_VERSION_INFO[0] < 3:
                 kwargs["fields"] = error.fields
+                kwargs["field_names"] = error.field_names
```
sloria, line 329: It might be worth checking if the version is <3.0.0rc1 so that 3.0.0 beta users' code won't be broken.

sloria: Latest commit does this.

lafrech:
Do we really need to maintain compatibility with all RC versions?

I thought once a pre-release version was out, all former pre-releases were considered obsolete.

sloria:
Eventually we can stop supporting all pre-release versions. But considering we only released 3.0.0rc1 a few days ago and many users are using 3.0 beta, I think we should support beta for this webargs release.

PR #332, opened 2018-12-01
BUG: create boolean check for location type allowing more options on marshmellow Field

attack68:
I have the following use-pattern, which involves transmitting an image file in some cases and in others a string as part of a form:

    'args': {"id": fields.Int(required=True), "image": fields.Field(location=('files', 'form'), allow_blank=True)}

This failed since the location input of Field was converted to a list before converting to a set in upon validation, here:

    def _validated_locations(self, locations):
        """Ensure that the given locations argument is valid.

        :raises: ValueError if a given locations includes an invalid location.
        """
        # The set difference between the given locations and the available locations
        # will be the set of invalid locations
        valid_locations = set(self.__location_map__.keys())
        given = set(locations)

 However the commit change permits this and is backwards compatible for regular single item string inputs.

### DIFF
--- a/webargs/core.py
+++ b/webargs/core.py
@@ -270,7 +270,10 @@ class Parser(object):
         """
         location = field.metadata.get("location")
         if location:
-            locations_to_check = self._validated_locations([location])
+            if isinstance(location, tuple) or isinstance(location, list):
+                locations_to_check = self._validated_locations(location)
+            else:
+                locations_to_check = self._validated_locations([location])
         else:
             locations_to_check = self._validated_locations(locations or self.locations)
 


webargs/core.py
```diff
@@ -270,7 +270,10 @@ def parse_arg(self, name, field, req, locations=None):
         """
         location = field.metadata.get("location")
         if location:
-            locations_to_check = self._validated_locations([location])
+            if isinstance(location, tuple) or isinstance(location, list):
```
sloria, line 273: This can be simplified to

```suggestion
            if isinstance(location, (list, tuple)):
```

sloria (changes requested):
Thanks for the PR!

I wouldn't consider this a bug fix since we don't currently support passing multiple locations to a field.

That said, I'm open to adding this. Can you please add a test for this and--if you have time--update any relevant docs?

attack68:
Sorry, have been very busy to look at this - will do if I get time. I was a bit put off by the extensive test failure report. It seemed to be a lot to go through and figure out what was going on from a newcomers perspective, for what I thought originally might be easy to slot it.

sloria:
@attack68 I understand. No rush to get to it if you don't have time.

The test failures are unrelated to this PR and have since been fixed. So if you rebase your branch with `dev` and force-push, the tests should pass.

sloria:
Closing for now, since this has gone stale. Feel free to re-open a PR if you're still interested in this.

PR #334, opened 2018-12-24
Return error response for invalid JSON payloads

### LINKED ISSUE #329: Throwing error on invalid JSON string…
_This seems too big to be an oversight, so apologies in advance if this is a design choice and it's been covered before (I looked in the issues and didn't see anything):_

When sending an invalid JSON string, eg (note the stray comma):
```
curl -X PUT "http://localhost:5000/bar/0001" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"foo\": \"bar\" , }"
```

… it seems that parsers are expected to simply return an empty dict `{}`, while silencing any error. Eg `flask_parser`:
```
        # Fail silently so that the webargs parser can handle the error
        if hasattr(req, "get_json"):
            # Flask >= 0.10.x
            json_data = req.get_json(force=force, silent=True)
        else:
            # Flask <= 0.9.x
            json_data = req.json
        if json_data is None:
            return core.missing
```

Despite what the comment implies, I don't think there is any way later on to distinguish between a real missing value, and one resulting from an invalid JSON string…

As expected, assuming an empty dict is compatible with Schema validation options, no error is thrown (and adding an error handler does not seem like it would help).

Am I missing something? Shouldn't JSON parsing errors be caught and trigger a 400 or 422 automatically?

sloria:
Previously, invalid JSON payloads were allowed to pass silently.
With this change, a HTTPError with status code 400 and a payload of `{"json": ["Invalid JSON body."]}` is raised instead.

close #329

Note: Adds simplejson as a dependency on Python 2.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,10 +4,9 @@ Changelog
 5.0.0 (unreleased)
 ******************
 
-* *Backwards-incompatible*: A ``JSONDecodeError`` is raised when an
-  invalid JSON payload is passed. This allows users to return a proper
-  error message for invalid JSON payloads instead of passing silently
-  (:issue:`329`). Thanks :user:`zedrdave` for reporting.
+* *Backwards-incompatible*: A 400 HTTPError is raised when an
+  invalid JSON payload is passed.  (:issue:`329`). 
+  Thanks :user:`zedrdave` for reporting.
 
 Other changes:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,11 @@ Changelog
   error message for invalid JSON payloads instead of passing silently
   (:issue:`329`). Thanks :user:`zedrdave` for reporting.
 
+Other changes:
+
+* `simplejson <https://pypi.org/project/simplejson/>`_ is now a dependency.
+  This ensures consistency of behavior across Python 2 and 3.
+
 4.3.0 (2018-12-30)
 ******************
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
+5.0.0 (unreleased)
+******************
+
+* *Backwards-incompatible*: A ``JSONDecodeError`` is raised when an
+  invalid JSON payload is passed. This allows users to return a proper
+  error message for invalid JSON payloads instead of passing silently
+  (:issue:`329`). Thanks :user:`zedrdave` for reporting.
+
 4.3.0 (2018-12-30)
 ******************
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,8 @@ Changelog
 
 Other changes:
 
-* `simplejson <https://pypi.org/project/simplejson/>`_ is now a dependency.
+* `simplejson <https://pypi.org/project/simplejson/>`_ is now a required
+  dependency on Python 2 (:pr:`334`).
   This ensures consistency of behavior across Python 2 and 3.
 
 4.3.0 (2018-12-30)
--- a/examples/bottle_example.py
+++ b/examples/bottle_example.py
@@ -12,7 +12,7 @@ Try the following with httpie (a cURL-like utility, http://httpie.org):
     $ http POST :5001/dateadd value=1973-04-10 addend=63
     $ http POST :5001/dateadd value=2014-10-23 addend=525600 unit=minutes
 """
-import json
+import simplejson as json
 import datetime as dt
 
 from bottle import route, run, error, response, HTTPResponse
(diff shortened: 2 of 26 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,21 @@
...
+Features:
+
+* *Backwards-incompatible*: A ``JSONDecodeError`` is raised when an
+  invalid JSON payload is passed. This allows users to return a proper
+  error message for invalid JSON payloads instead of passing silently
+  (:issue:`329`). Thanks :user:`zedrdave` for reporting.
+
+Other changes:
+
+* `simplejson <https://pypi.org/project/simplejson/>`_ is now a dependency.
```
lafrech: Can you elaborate on this?

https://pypi.org/project/simplejson/

> simplejson is the externally maintained development version of the json library included with Python 2.6 and Python 3.0, but maintains backwards compatibility with Python 2.5.

I didn't test, but the README says it is included in 2.6+ and 3.0+. How come we need it in the first place?

sloria: We need simplejson's backport of `json.JSONDecodeError`, which doesn't exist on Python 2.7.

lafrech: Ah, OK.

Do you think we could make the requirement conditional, depending on the Python version?

Or add a comment to remember to remove it when dropping Python 2.x?

No big deal otherwise.

sloria: Hm, conditionally requiring simplejson on Python 2 won't work OOB because

```python
In [1]: import simplejson

In [2]: import json

In [3]: simplejson.JSONDecodeError is json.JSONDecodeError
Out[3]: False
```

I suppose we could do something like

```python
# webargs/json.py

try:
    import simplejson as json
except ImportError:
    import json


# in parser modules
from webargs.core import parse_json
from webargs.json import json

...
try:
    parse_json(data)
except json.JSONDecodeError as err:
    if err.doc == "":
       ....
```

I'd be fine with that, I suppose.

sloria: Actually, we could just import json from `webargs.core`:

```python
# in parser modules
from webargs.core import parse_json, json
...

try:
    parse_json(data)
except json.JSONDecodeError as err:
    if err.doc == "":
       ....
```

sloria: OK, updated the code so that simplejson is only required on Python 2.

lafrech:
I didn't inspect thoroughly so this is a bit of a lazy question, but wouldn't it be more in line with the validation error triggering a 422 if the parser aborted with 400 rather than raising an exception for the user to catch?

I find it a bit dodgy that we raise an exception that will trigger a 500 out of the box. Even with a major release and a warning.

I can live with that change. I'll just need to add a handler as you suggest, but is there a reason not to return a 400 error directly? The message could be personalized with a parser attribute if that's a concern. And the formatting can be done by catching the `HTTPError` just like with the 422. At least in the use case I'm familiar with (Flask + flask_rest_api).

sloria:
You raise a fair point. I think I went the 500 route for consistency with the current behavior of aiohttpparser and for configurability. But in retrospect, it's more user-friendly to raise an `HTTPError`. I'll try this out when I have some time.

sloria:
Rebased and modified so that a 400 response is returned with the payload:

```json
{"json": ["Invalid JSON body."]}
```

A `handle_invalid_json_error` hook was added to the relevant parsers to make this behavior configurable.

examples/bottle_example.py
```diff
@@ -58,6 +58,7 @@ def dateadd(value, addend, unit):
 
 
 # Return validation errors as JSON
+@error(400)
 @error(422)
 def error422(err):
```
lafrech: Should this be renamed (remove 422 in the name)?

setup.py
```diff
@@ -1,8 +1,12 @@
 # -*- coding: utf-8 -*-
+import sys
 import re
 from setuptools import setup, find_packages
 
 INSTALL_REQUIRES = ["marshmallow>=2.15.2"]
+if sys.version_info[0] < 3:
+    INSTALL_REQUIRES.append("simplejson")
```
lafrech: Add a minimal version? "simplejson>=x.x.x"?

tests/apps/bottle_app.py
```diff
@@ -115,7 +115,8 @@ def echo_nested_many():
     return parser.parse(args)
 
 
-@error(422)
+@app.error(400)
+@app.error(422)
 def handle_422(err):
```
lafrech: Likewise, 422 in the name.

lafrech:
Looks good to me! :+1:

examples/bottle_example.py
sloria: Good catch. Fixed.

lafrech:
Great. This is good to go.

PR #338, opened 2018-12-24
Deprecate status_code and headers params of ValidationErrors

### LINKED ISSUE #336: Deprecate status_code and headers argument to ValidationError
As an incremental step towards #327, raise a deprecation warning when `status_code` is passed to `ValidationError`.

sloria:
* `status_code` and `headers` kwargs are deprecated (close #336)
* Add `error_status_code` and `error_headers`:
  See https://github.com/sloria/webargs/issues/327#issuecomment-449669326

```python
def validate_foo(val):  # Before
    ...
    raise ValidationError("Error!", status_code=400)

@use_args({"foo": fields.Str(validate=validate_foo)})
def my_view(args):
    ...

def validate_foo(val):  # After
    ...
    raise ValidationError("Error!")

@use_args(
    {"foo": fields.Str(validate=validate_foo)},
    error_status_code=400
)
def my_view(args):
    ...
```

* error_handler functions and handle_error methods now take
  error_status_code and error_headers arguments. Backwards
  compatibility was maintained by checking the the number of
  arguments in the signatures.

```python
@parser.error_handler
def handle_error(error, req):  # Before
    raise CustomError(error.messages)

@parser.error_handler
def handle_error(error, req, status_code, headers):  # After
    raise CustomError(error.messages)

@parser.error_handler
def handle_error(error, **kwargs):  # Also works
    raise CustomError(error.messages)
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,53 @@ Features:
 
 * Add ``force_all`` argument to ``use_args`` and ``use_kwargs``
   (:issue:`252`, :issue:`307`). Thanks :user:`piroux` for reporting.
+* The ``status_code`` and ``headers`` arguments to ``ValidationError``
+  are deprecated. Pass ``error_status_code`` and ``error_headers`` to
+  `Parser.parse`, `Parser.use_args`, and `Parser.use_kwargs` instead.
+  (:issue:`327`, :issue:`336`).
+* Custom error handlers receive ``error_status_code`` and ``error_headers`` arguments.
+  (:issue:`327`).
+
+.. code-block:: python
+
+    # <4.2.0
+    @parser.error_handler
+    def handle_error(error, req, schema):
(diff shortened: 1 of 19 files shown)


lafrech (approved):

lafrech:
LGTM.

I'd have dropped the whole status_code/headers feature altogether as I said in https://github.com/sloria/webargs/issues/327, but if that serves legit use cases, I'm fine with it. This PR impacts a lot of files but that's mainly because of examples and deprecation warnings overhead. In practice, it's not so much API surface.

In 5.0, we'll be able to get rid of the `ValidationError` subclass to use the vanilla one from marshmallow.

sloria:
Agreed. Though maybe we should still make it importable from webargs, like with fields.

```python
from webargs.core import ValidationError # just aliases marshmallow.ValidationError
```

I don't feel too strongly about it.

PR #376, opened 2019-03-15
Make schema class configurable

ThiefMaster:


sloria (approved):
Looks good! Thanks for the PR. I just renamed "schema_cls" to "schema_class" because there's no reason to abbreviate "cls" in this case, and it's more consistent with, e.g. `OPTIONS_CLASS`.

sloria:
For anyone finding this... there's a proposal for adding this "dict-to-schema" behavior into marshmallow core: https://github.com/marshmallow-code/marshmallow/issues/1312

PR #378, opened 2019-03-18
Use a proper name for implicit schema classes

ThiefMaster:
Otherwise you end up with ugly reprs like `<class 'marshmallow.schema.'>`

### DIFF
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -64,7 +64,7 @@ def dict2schema(dct, schema_class=ma.Schema):
             register = False
 
     attrs["Meta"] = Meta
-    return type(str(""), (schema_class,), attrs)
+    return type(str("WebargsDictSchema"), (schema_class,), attrs)
 
 
 def is_multiple(field):


sloria:
The reason we use an empty string is to avoid adding the class to the internal class registry in marshmallow 2 (in 3.0, this is taken care of with `register = False`.

I *think* this will end up adding multiple entries in the registry, but I'm not sure. @ThiefMaster would you mind checking this? I can also check it later when I'm off hours.

ThiefMaster:
Ah, didn't think about that. Just noticed the weird repr while debugging something in my app.

sloria:
Yeah, this will be an issue on marshmallow versions older than 2.16.3.

```python
In [1]: from marshmallow import class_registry

In [2]: from webargs.core import dict2schema

In [3]: dict2schema({})
Out[3]: marshmallow.schema.WebargsDictSchema

In [4]: dict2schema({})
Out[4]: marshmallow.schema.WebargsDictSchema

In [5]: dict2schema({})
Out[5]: marshmallow.schema.WebargsDictSchema

In [6]: class_registry._registry
Out[6]:
{'Schema': [marshmallow.schema.Schema],
 'marshmallow.schema.Schema': [marshmallow.schema.Schema],
 'WebargsDictSchema': [marshmallow.schema.WebargsDictSchema],
 'marshmallow.schema.WebargsDictSchema': [marshmallow.schema.WebargsDictSchema,
  marshmallow.schema.WebargsDictSchema,
  marshmallow.schema.WebargsDictSchema]}
```

Closing this for now. We can do this when we bump up the minimum marshmallow version.

PR #379, opened 2019-03-22
add aiohttp path location

zhenhua32:
alias to match_info


sloria:
I've no strong objection to this, but why do you think it's necessary? Thus far, we've tried to use the framework specific terminology for path params: "view args" in Flask, "matchdict" in Pyramid, "match info" in aiohttp.

zhenhua32:
compatible with OpenAPI , in OpenAPI match_info is normally called as path

sloria:
Fine by me. We'll probably want to make this consistent across all the parsers, but no reason to block merging. 🚢 🇮🇹

PR #395, opened 2019-06-18
Fix marshmallow 3.0.0rc7 compatibility

lafrech:
Since 3.0.0rc7, decorated methods should accept kwargs.

Also, since 3.0.0b19, `Field._deserialize` should accept kwargs (https://github.com/marshmallow-code/marshmallow/pull/1007). This was not done in webargs' custom fields, but no test was broken so we didn't see it.

This PR addresses all those issues.


sloria:
Ah good catch. Thanks

PR #407, opened 2019-07-24
Document parsing lists in query strings; remove duck typing

### LINKED ISSUE #406: Improve documentation on fields.DelimitedList vs fields.List
The docs for the fields.DelimitedList say it "can load from either a list or a delimited string" but I cannot get it to load from a list. `It's possible I'm doing something wrong` but When I swap out `fields.DelimitedList` for `fields.List` everything works as expected. I think either the docs should be changed, or we should fix the implementation of `DelimitedList` so that it works as expected.

sloria:
close #406

### DIFF
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -65,7 +65,11 @@ def dict2schema(dct, schema_class=ma.Schema):
 
 def is_multiple(field):
     """Return whether or not `field` handles repeated/multi-value arguments."""
-    return isinstance(field, ma.fields.List) and not hasattr(field, "delimiter")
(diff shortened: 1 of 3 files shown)


src/webargs/core.py
```diff
@@ -65,7 +65,11 @@ class Meta(object):
 
 def is_multiple(field):
     """Return whether or not `field` handles repeated/multi-value arguments."""
-    return isinstance(field, ma.fields.List) and not hasattr(field, "delimiter")
```
sloria, line 68: We probably did this to avoid the circular import. Was a bad decision in retrospect.

Issue #267, opened 2018-08-08
Leave logic of deciding on extra args to schema


tuukkamustonen:
Now that marshmallow 3.x is in beta and comes with configurable behavior on extra args, how about just including all fields in `Parser._parse_request()` and passing them on to  `Schema.load()`? It would be up to passed Schema to determine what to do with them.

The effect would be that:

- With `RAISE`,  extra fields would be denied.
- With `INCLUDE`, extra fields would be allowed and passed on.
- With `EXCLUDE`, extra fields would be ignored.

This would allow something like:

```python
@use_args(HeaderSchema(unknown=EXCLUDE), locations=('headers', ))
@use_args({'action': fields.Str()}, locations=('json',))  # default is RAISE
```

Actually, I don't think this even needs marshmallow `3.x`. Why not just always pass all fields from `_parse_request()` to `Schema.load()`? On `2.x` you would need custom `@validates_schema` implementation in addition, but on `3.x` you would not.

Maybe I'm missing something...

sloria:
Thanks for the suggestion @tuukkamustonen ; I think it's a good one. I don't think there's a [good] reason for `_parse_request` to filter on declared fields.

Would you be up for sending a PR?

tuukkamustonen:
I can try to come up with something (hopefully within a few days or so).

Edit: Turns out I bumped into other tickets/issues and this got delayed. Still hoping I (or anyone) can dig up time for it!

lafrech:
Related to https://github.com/sloria/webargs/issues/173.

toonalbers:
Hi all,

Finding all fields seems less trivial than I had hoped, but I have tried to think up a solution with minimal changes to the existing code. My idea was to have each parser provide a list of fields it knows at each location. Then `_parse_request` would still do what it currently does, but afterwards also parses values in the other discovered fields and adds them to the result.

Benefits of this approach:
- No need to change large parts of the webargs codebase
- Backwards compatible if we enable it through some option

Downsides to this:
- I'm not sure if/how undeclared _nested_ arguments would work since I use `ma.fields.Raw` as a default.
- Using  `unknown=RAISE` would break when using `locations=('headers',)` as you would get _all_ included HTTP headers.

Would [something like this](https://github.com/toonalbers/webargs/commit/b433ea59656f001e5cb642fd287bb2697c2ac373) be interesting to pursue? (I did not finish or test it.)

sirosen:
[...]

PR #410, opened 2019-08-04
Pass all args to schemas, but default to unknown=EXCLUDE (work in progress)

sirosen:
This is a work in progress / demonstration of a solution for #267 .
It's based on the suggestion in that thread (big thanks to @toonalbers for that!).

I did two versions of this and have kept it in separate commits so that both can be reviewed if desirable. I prefer the latter form and will squash the work if it's going to be accepted in a form even remotely like this.

In the first version, if Parser.pass_all_args is set (it defaults to false for backwards compatibility), then argument parsing becomes a two step process.
First, args are parsed based on the schema fields, as today. Then, second, they are all treated as `marshmallow.fields.Raw`, parsed by webargs, and the result of that parsing is passed to the schema for validation.
If Parser.pass_all_args is set on marshmallow versions < 3, it is a ValueError.

As a refinement on this, I successfully swapped out Parser.pass_all_args for Parser.unknown, which sets the value to pass for `Schema.load(unknown=...)`. On marshmallow v3, it defaults to `EXCLUDE`, which makes this backwards-compatible. Callers may pass `None` to send no value and obey the schema default. All arguments are *always* passed to the schema, and we're just using the very `unknown=...` behavior we're trying to expose to maintain backwards compatibility.

(Note: this is arguable backwards *incompatible* because certain usages with strict schema validators were passing and will start to fail.)

In both versions, in order to collect arguments to pass to the schema, this adds a new function required on all parsers, `get_args_by_location`, which collects all argument names, typically dict keys, categorized by location name.
This is only implemented here in the flask parser, as a proof of concept.

There's a lot more work that I would need to do for this to be ready to merge. Namely:
- many more tests
- support in all built-in parsers for `get_args_by_location`

However, I don't want to put in the labor to do the kind of testing I think this change needs without gettings eyes on this work from a webargs maintainer and confirmation that this whole approach is okay. If it is, I'll happily put together the necessary work to make all of the parsers handle this and do a lot more testing.

### DIFF
--- a/src/webargs/aiohttpparser.py
+++ b/src/webargs/aiohttpparser.py
@@ -78,19 +78,13 @@ class AIOHTTPParser(AsyncParser):
         **core.Parser.__location_map__
     )
 
-    def parse_querystring(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a querystring value from the request."""
-        return core.get_value(req.query, name, field)
-
-    async def parse_form(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a form value from the request."""
+    async def _load_post_data(self, req: Request) -> typing.Any:
         post_data = self._cache.get("post")
         if post_data is None:
-            self._cache["post"] = await req.post()
-        return core.get_value(self._cache["post"], name, field)
+            self._cache["post"] = post_data = await req.post()
+        return post_data
 
-    async def parse_json(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a json value from the request."""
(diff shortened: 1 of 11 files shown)


sirosen:
I got impatient and just added implementations of `get_args_by_location` for all parsers. So the only thing that I think this really needs is an expansion of the testsuite, which I'll try to carve out time to work on in coming weeks.

sloria:
Thanks for the PR. Apologies for the delay.. I'll need to carve out time to give this a proper review. I've not forgotten about it!

sloria:
closing for now, since you're carrying on with #420

PR #417, opened 2019-08-19
Ensure a compatible version of marshmallow is used

jensgustafsson:
`marshmallow` 3.0.0 has made a breaking change since `marshmallow.compat` no longer exists. That module is used in several locations.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import sys
 import re
 from setuptools import setup, find_packages
 
-INSTALL_REQUIRES = ["marshmallow>=2.15.2"]
+INSTALL_REQUIRES = ["marshmallow>=2.15.2,<3.0.0"]
 if sys.version_info[0] < 3:
     INSTALL_REQUIRES.append("simplejson>=2.1.0")
 


lafrech:
webargs is compatible with both marshmallow 2 and 3.

webargs + MA3 issues should be reported here and fixed.

Do you have a specific issue when using webargs + MA3?

jensgustafsson:
Here's one example when version 3 is not compatible:

When `webargs.core` is imported, it gets an `ModuleNotFoundError`: 

```
ModuleNotFoundError: No module named 'marshmallow.compat'
```

`webargs/core.py:17`:

```
import marshmallow as ma
from marshmallow.compat import iteritems  # <--- marshmallow.compat does not exist
```

lafrech:
I'm pretty sure this was fixed in https://github.com/marshmallow-code/webargs/commit/27f53de9d385ccaa802775f86cd64034393d0102.

Can you reproduce with the latest webargs version?

jensgustafsson:
My bad! I can see that I mixed up `marshmallow.compat` with `webargs.compat` when I looked in the code.

Apparently I'm running webargs 4.x. Upgrading to webargs 5.x should make my issue go away. Anyhow, this fix should probably be applied to the latest 4.x version. I'm closing this issue, but the issue still remains in webargs 4.x.

lafrech:
We don't maintain several branches. Only latest stable branch is maintained. Users should move to 5.x.

Issue #268, opened 2018-08-08
Inconsistent behavior in passing request fields to Schema.load (many=True vs many=False)


tuukkamustonen:
With `4.0.0` (and marshmallow `3.0.0b13`):

```python
class Sch(Schema):
    foo = fields.Str()

parser = FlaskParser()

req = Mock()
req.mimetype = 'application/json'

req.get_json = lambda *args, **kwargs: [{'extra': 1}]
with pytest.raises(UnprocessableEntity) as e:
    parser.parse(Foo(many=True), req=req, locations=('json', ))
assert e.value.exc.messages == {0: {'extra': ['Unknown field.']}}

req.get_json = lambda *args, **kwargs: {'extra': 1}
parser.parse(Foo(), req=req, locations=('json', ))
```

So the with `many=True` the extra key is actually passed to `Schema.load()` (which can then raise `ValidationError` because `unknown` defaults to `RAISE`) but with `many=False` it is passed onwards. I think it's confusing.

#267 would fix this.

lafrech:
@tuukkamustonen is this fixed in webargs 6 beta?

tuukkamustonen:
@lafrech I haven't been able to follow the project lately, but as the referenced ticket #267 has been fixed and merged, I would assume so.

Feel free to close, if you think this is fixed, we can re-open if it's not...

lafrech:
Fixed.

```py
from werkzeug.exceptions import UnprocessableEntity
from marshmallow import Schema, fields
from webargs.flaskparser import FlaskParser
from unittest.mock import Mock


class Foo(Schema):
    foo = fields.Str()

parser = FlaskParser()

req = Mock()
req.mimetype = 'application/json'

req.get_data = lambda *args, **kwargs: '[{"extra": 1}]'

# Both raise UnprocessableEntity
try:
    parser.parse(Foo(many=True), req=req, location='json')
except UnprocessableEntity:
    print('OK')
try:
    parser.parse(Foo(), req=req, location='json')
except UnprocessableEntity:
    print('OK')
```

tuukkamustonen:
This indeed looks like fixed now. Thanks for checking!

tuukkamustonen:
Just a minor "correction":

The latter case, with `many=False` actually failed with just `{"extra": 1}` not `[{"extra": 1}]` (as in your code above).

But both those now raise error properly, so yeah it's fixed.

Issue #164, opened 2017-05-03
Non-iterable values get passed as [] to @pre_load @validates_schema et al.


tuukkamustonen:
With:

```python
class Foo(StrictSchema):
    foo = fields.Str()

parser = FlaskParser()
    
req = Mock()
req.mimetype = 'application/json'
req.get_json = lambda *_, **__: {'foo': 1}  # or any non-iterable

parser.parse(Foo(strict=True, many=True), req=req, locations=('json', ))
```

`@pre_load(pass_original=True)` `@validate_schema(pass_original=True)` get `[]` as original value, instead of `{'foo': 1}`.

This makes it impossible to record errors about invalid base type / extra fields.

This happens for any non-iterable non-string non-mapping value.

The code that's related to this is in `get_value()` at https://github.com/sloria/webargs/blob/dev/webargs/core.py#L130-L132. Not yet sure what all that processing really does, but could that function return the data as is...?

tuukkamustonen:
Hmm, I think this is because of #267 (`_parse_request()` skipping on extra keys / invalid data).

decaz:
@tuukkamustonen @sloria I've found the test which I think related to the issue:
https://github.com/marshmallow-code/webargs/blob/cfe9c4aad5c86c82d0e829577ce23e96f3f2963d/src/webargs/testing.py#L96-L97

The question is why invalid input type is implicitly converted into an empty list? Shouldn't the method return a response with the error "Invalid input type" ?

lafrech:
Is this fixed in webargs 6 beta?

tuukkamustonen:
Tried to check it, but so many things have changed and I don't have time to go deeper into it right now.

lafrech:
With

```py
from marshmallow import Schema, fields
from webargs.flaskparser import FlaskParser
from unittest.mock import Mock


class Foo(Schema):
    foo = fields.Str()

parser = FlaskParser()

req = Mock()
req.mimetype = 'application/json'
req.get_data = lambda *_, **__: '{"foo": 1}'  # or any non-iterable

parser.parse(Foo(many=True), req=req, location='json')
```

I get

    marshmallow.exceptions.ValidationError: {'_schema': ['Invalid input type.']}

    werkzeug.exceptions.UnprocessableEntity: 422 Unprocessable Entity: The request was well-formed but was unable to be followed due to semantic errors.


I guess this means it is fixed.

tuukkamustonen:
[...]

Issue #366, opened 2019-02-01
AttributeError: module 'typing' has no attribute 'NoReturn' with Python 3.5.3


lafrech:
I get this error when running the tests with Python 3.5.3.

```
tests/test_py3/test_aiohttpparser_async_functions.py:6: in <module>
    from webargs.aiohttpparser import parser, use_args, use_kwargs
webargs/aiohttpparser.py:72: in <module>
    class AIOHTTPParser(AsyncParser):
webargs/aiohttpparser.py:148: in AIOHTTPParser
    ) -> typing.NoReturn:
E   AttributeError: module 'typing' has no attribute 'NoReturn'
```

The docs say [`typing.NoReturn`](https://docs.python.org/3/library/typing.html#typing.NoReturn) was added in 3.6.5. However, [the tests pass on Travis](https://travis-ci.org/marshmallow-code/webargs/jobs/486701760) with Python 3.5.6.

sloria:
Looks like this was a bug until 3.5.4rc1 : https://bugs.python.org/issue28556 .

@lafrech Can you upgrade your Python version?

lafrech:
Not really. https://packages.debian.org/search?keywords=python3

I could use pyenv to develop and run the tests. Still, people using Debian stable or any 3.5.3- version can't run the asyncio part of the code. I mean it is not only about me running the tests but also about anyone willing to run an application using the lib.

But as long as no one complains, maybe we can live with this until 3.6 or 3.7 is standard.

sloria:
I see. I'd prefer not to support old patch versions Python, but we can live with removing the `typing.NoReturn` usages for now, at least until we decide to drop 3.5 support.

PR #420, opened 2019-09-04
Only allow users to specify a single location per parse call & pass full location data to schema.load

### LINKED ISSUE #419: RFC: Allow a single location for each field
See discussion in https://github.com/marshmallow-code/webargs/issues/267.

The idea is that the Schema would be passed the whole data from a location, so there couldn't be fields with different location in the same Schema.

- Drop `location` argument to `Field`
- Only accept a single `location` in `use_args`

This is an important breaking change.

It breaks the default use case, which is to iterate on all fields, and for each field search in `DEFAULT_LOCATIONS` (by default: query string, form, then json).

We could keep a somewhat similar behaviour by trying to load the whole data from query string, then form, then json. The only added restriction is that Schema couldn't contain fields from mixed locations.

I'm not convinced with the multi-location feature, so we could even

- Replace `DEFAULT_LOCATIONS` with `DEFAULT_LOCATION`

The user could subclass the parser to change the default location (in our framework, we use json by default), and he would need to specify the location explicitly when it is not the default.

I think all this changes allow for better documentation (e.g. in apispec).

sirosen:
This is still incomplete, but core and flaskparser tests are now passing. No work is (yet) included on other concrete parsers, but I fixed the base asyncparser to make mypy linting pass.

Resolves #419, #267, #164, #268 
Obsoletes, and therefore closes, #410

I've updated the changelog with a basic "statement of intent". As I work on this, I will try to maintain it as an accurate record of how this change should look to the outside world. Obviously, narrative docs also need to be updated with relevant info.

Highlights / greatest hits:
- Marks this as an unreleased v6.0.0 in the changelog
- Replace `DEFAULT_LOCATIONS` with `DEFAULT_LOCATION="json"`
- Replace `parse_{location}` with `location_load_{location}`
- Replace `locations=<iterable>` with `location=<str>`
- Remove `locations=...` from Fields
- Remove `Parser.parse_arg`
- Remove `webargs.core.get_value`
- Add `webargs.multidictproxy.MultiDictProxy` which has `get_value`-like behavior and proxies various dict-like objects
- Reverse the decision made in #297 with [a comment in tests explaining this change](https://github.com/sirosen/webargs/blob/90c337278677aaca603b1b8f71aa914ceaad8dca/src/webargs/testing.py#L205-L213) and a bit of [similar explanation in the commit message](
https://github.com/marshmallow-code/webargs/commit/e59963f3994e26ee108226d52536af943ec9b961)
- Numerous fixes to tests to be more explicit about which location is being used (since you can't trivially have a single location `/echo` which loads data from a bunch of locations)
- As a side-effect of the above changes, certain inputs which were previously ignored are treated as errors (e.g. `1` for a JSON body will now parse and be passed literally to the schema, which will fail because `1` is not a valid input for `schema.load`). Changes to the tests serve as a somewhat-readable list of behavioral changes.

To try to keep this from sprawling *even more*, my intent is to address some items mentioned in #419 either in follow-up PRs or as part of this only if we're happy with this direction (and agree on those items). In particular: the `json_or_form` location I proposed and nested `use_args` usage and error collection.

@sloria, @lafrech, my biggest question is: do you like where this is going? Should I just hammer out all of the other parsers?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,7 +16,7 @@ Refactoring:
 
 * *Backwards-incompatible*: Schema fields may not specify a location any
   longer, and `Parser.use_args` and `Parser.use_kwargs` now accept `location`
-  (singular) instead of `locations` plural. Instead of using a single field or
+  (singular) instead of `locations` (plural). Instead of using a single field or
   schema with multiple `locations`, users are recommended to make multiple
   calls to `use_args` or `use_kwargs` with a distinct schema per location. For
   example, code should be rewritten like this:
@@ -24,26 +24,28 @@ Refactoring:
 .. code-block:: python
 
     # under webargs v5
-    class CompoundSchema:
-        q1 = ma.fields.Int(location="query")
-        q2 = ma.fields.Int(location="query")
-        h1 = ma.fields.Int(location="headers")
-
-    @parser.use_args(CompoundSchema(), locations=("query", "headers"))
+    @parser.use_args(
+        {
+            "q1": ma.fields.Int(location="query"),
+            "q2": ma.fields.Int(location="query"),
+            "h1": ma.fields.Int(location="headers"),
+        },
+        locations=("query", "headers"),
+    )
     def foo(q1, q2, h1):
         ...
+
 
     # should be split up like so under webargs v6
-    class QueryParamSchema:
-        q1 = ma.fields.Int()
-        q2 = ma.fields.Int()
-    class HeaderSchema:
-        h1 = ma.fields.Int()
-
-    @parser.use_args(QueryParamSchema(), location="query")
-    @parser.use_args(HeaderSchema(), location="headers")
+    @parser.use_args({"q1": ma.fields.Int(), "q2": ma.fields.Int()}, location="query")
+    @parser.use_args({"h1": ma.fields.Int()}, location="headers")
     def foo(q1, q2, h1):
         ...
+
+* The `location_handler` decorator has been removed and replaced with
+  `location_loader`. `location_loader` serves the same purpose (letting you
+  write custom hooks for loading data) but its expected method signature is
+  different. See the docs on `location_loader` for proper usage.
 
 5.5.2 (2019-10-06)
 ******************
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,50 @@
 Changelog
 ---------
 
+6.0.0 (unreleased)
+******************
+
+Features:
+
+* *Backwards-incompatible*: Schemas will now load all data from a location, not
+  only data specified by fields. As a result, schemas with validators which
+  examine the full input data may change in behavior. The `unknown` parameter
+  on schemas may be used to alter this. For example,
+  `unknown=marshmallow.EXCLUDE` will produce behavior similar to webargs v5
+
+Refactoring:
+
+* *Backwards-incompatible*: Schema fields may not specify a location any
+  longer, and `Parser.use_args` and `Parser.use_kwargs` now accept `location`
+  (singular) instead of `locations` plural. Instead of using a single field or
+  schema with multiple `locations`, users are recommended to make multiple
+  calls to `use_args` or `use_kwargs` with a distinct schema per location. For
+  example, code should be rewritten like this:
+
+.. code-block:: python
+
+    # under webargs v5
+    class CompoundSchema:
+        q1 = ma.fields.Int(location="query")
+        q2 = ma.fields.Int(location="query")
+        h1 = ma.fields.Int(location="headers")
+
+    @parser.use_args(CompoundSchema(), locations=("query", "headers"))
+    def foo(q1, q2, h1):
+        ...
+
+    # should be split up like so under webargs v6
+    class QueryParamSchema:
+        q1 = ma.fields.Int()
+        q2 = ma.fields.Int()
+    class HeaderSchema:
+        h1 = ma.fields.Int()
+
+    @parser.use_args(QueryParamSchema(), location="query")
+    @parser.use_args(HeaderSchema(), location="headers")
+    def foo(q1, q2, h1):
+        ...
+
 5.5.2 (2019-10-06)
 ******************
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,12 @@ Features:
   on schemas may be used to alter this. For example,
   `unknown=marshmallow.EXCLUDE` will produce behavior similar to webargs v5
 
+Bug fixes:
+
+* *Backwards-incompatible*: all parsers now require the Content-Type to be set
+  correctly when processing JSON request bodies. This impacts ``DjangoParser``,
+  ``FalconParser``, ``FlaskParser``, and ``PyramidParser``
+
 Refactoring:
 
 * *Backwards-incompatible*: Schema fields may not specify a location any
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -27,7 +27,7 @@ Arguments are specified as a dictionary of name -> :class:`Field <marshmallow.fi
         # or you want to rename a key
         "user_type": fields.Str(load_from="user-type"),
         # OR, on marshmallow 3
-        # "content_type": fields.Str(data_key="user-type"),
+        # "user_type": fields.Str(data_key="user-type"),
     }
 
 .. note::
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -23,17 +23,11 @@ Arguments are specified as a dictionary of name -> :class:`Field <marshmallow.fi
         "nickname": fields.List(fields.Str()),
         # Delimited list, e.g. "/?languages=python,javascript"
         "languages": fields.DelimitedList(fields.Str()),
-        # When you know where an argument should be parsed from
-        "active": fields.Bool(location="query"),
         # When value is keyed on a variable-unsafe name
         # or you want to rename a key
-        "content_type": fields.Str(load_from="Content-Type", location="headers"),
+        "user_type": fields.Str(load_from="user-type"),
         # OR, on marshmallow 3
-        # "content_type": fields.Str(data_key="Content-Type", location="headers"),
-        # File uploads
-        "profile_image": fields.Field(
-            location="files", validate=lambda f: f.mimetype in ["image/jpeg", "image/png"]
-        ),
+        # "content_type": fields.Str(data_key="user-type"),
     }
 
 .. note::
@@ -105,12 +99,12 @@ As an alternative to `Parser.parse`, you can decorate your view with :meth:`use_
 Request "Locations"
 -------------------
 
-By default, webargs will search for arguments from the URL query string (e.g. ``"/?name=foo"``), form data, and JSON data (in that order). You can explicitly specify which locations to search, like so:
+By default, webargs will search for arguments from the request body as JSON. You can specify a different location from which to load data like so:
 
 .. code-block:: python
 
     @app.route("/register")
-    @use_args(user_args, locations=("json", "form"))
+    @use_args(user_args, location="form")
     def register(args):
         return "registration page"
 
@@ -243,7 +237,7 @@ Nesting Fields
 
 .. note::
 
-    By default, webargs only parses nested fields using the ``json`` request location. You can, however, :ref:`implement your own parser <custom-parsers>` to add nested field functionality to the other locations.
+    Of the default supported locations in webargs, only the ``json`` request location supports nested datastructures. You can, however, :ref:`implement your own data loader <custom-loaders>` to add nested field functionality to the other locations.
 
 Next Steps
 ----------
--- a/src/webargs/aiohttpparser.py
+++ b/src/webargs/aiohttpparser.py
@@ -89,6 +89,14 @@ class AIOHTTPParser(AsyncParser):
             self._cache["post"] = await req.post()
         return MultiDictProxy(self._cache["post"], schema)
 
+    async def load_json_or_form(
+        self, req: Request, schema: Schema
+    ) -> typing.Union[typing.Dict, MultiDictProxy]:
+        data = await self.load_json(req, schema)
+        if data is not core.missing:
+            return data
+        return await self.load_form(req, schema)
+
     async def load_json(self, req: Request, schema: Schema) -> typing.Dict:
         """Return a parsed json payload from the request."""
         json_data = self._cache.get("json")
--- a/src/webargs/aiohttpparser.py
+++ b/src/webargs/aiohttpparser.py
@@ -28,11 +28,11 @@ from aiohttp import web
 from aiohttp.web import Request
 from aiohttp import web_exceptions
 from marshmallow import Schema, ValidationError
-from marshmallow.fields import Field
 
 from webargs import core
 from webargs.core import json
 from webargs.asyncparser import AsyncParser
+from webargs.multidictproxy import MultiDictProxy
 
 
 def is_json_request(req: Request) -> bool:
@@ -73,24 +73,24 @@ class AIOHTTPParser(AsyncParser):
     """aiohttp request argument parser."""
 
     __location_map__ = dict(
-        match_info="parse_match_info",
-        path="parse_match_info",
+        match_info="load_match_info",
+        path="load_match_info",
         **core.Parser.__location_map__
     )
 
-    def parse_querystring(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a querystring value from the request."""
-        return core.get_value(req.query, name, field)
+    def load_querystring(self, req: Request, schema: Schema) -> MultiDictProxy:
+        """Return query params from the request as a MultiDictProxy."""
+        return MultiDictProxy(req.query, schema)
 
-    async def parse_form(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a form value from the request."""
+    async def load_form(self, req: Request, schema: Schema) -> MultiDictProxy:
+        """Return form values from the request as a MultiDictProxy."""
         post_data = self._cache.get("post")
         if post_data is None:
             self._cache["post"] = await req.post()
-        return core.get_value(self._cache["post"], name, field)
+        return MultiDictProxy(self._cache["post"], schema)
 
-    async def parse_json(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a json value from the request."""
+    async def load_json(self, req: Request, schema: Schema) -> typing.Dict:
+        """Return a parsed json payload from the request."""
         json_data = self._cache.get("json")
         if json_data is None:
             if not (req.body_exists and is_json_request(req)):
@@ -101,30 +101,30 @@ class AIOHTTPParser(AsyncParser):
                 if e.doc == "":
                     return core.missing
                 else:
-                    return self.handle_invalid_json_error(e, req)
+                    return self._handle_invalid_json_error(e, req)
             except UnicodeDecodeError as e:
-                return self.handle_invalid_json_error(e, req)
+                return self._handle_invalid_json_error(e, req)
 
             self._cache["json"] = json_data
-        return core.get_value(json_data, name, field, allow_many_nested=True)
+        return json_data
 
-    def parse_headers(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a value from the header data."""
-        return core.get_value(req.headers, name, field)
+    def load_headers(self, req: Request, schema: Schema) -> MultiDictProxy:
+        """Return headers from the request as a MultiDictProxy."""
+        return MultiDictProxy(req.headers, schema)
 
-    def parse_cookies(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a value from the cookiejar."""
-        return core.get_value(req.cookies, name, field)
+    def load_cookies(self, req: Request, schema: Schema) -> MultiDictProxy:
+        """Return cookies from the request as a MultiDictProxy."""
+        return MultiDictProxy(req.cookies, schema)
 
-    def parse_files(self, req: Request, name: str, field: Field) -> None:
+    def load_files(self, req: Request, schema: Schema) -> "typing.NoReturn":
         raise NotImplementedError(
-            "parse_files is not implemented. You may be able to use parse_form for "
+            "load_files is not implemented. You may be able to use load_form for "
             "parsing upload data."
         )
 
-    def parse_match_info(self, req: Request, name: str, field: Field) -> typing.Any:
-        """Pull a value from the request's ``match_info``."""
-        return core.get_value(req.match_info, name, field)
+    def load_match_info(self, req: Request, schema: Schema) -> typing.Mapping:
+        """Load the request's ``match_info``."""
+        return req.match_info
 
     def get_request_from_view_args(
         self, view: typing.Callable, args: typing.Iterable, kwargs: typing.Mapping
@@ -140,7 +140,8 @@ class AIOHTTPParser(AsyncParser):
             elif isinstance(arg, web.View):
                 req = arg.request
                 break
-        assert isinstance(req, web.Request), "Request argument not found for handler"
+        if not isinstance(req, web.Request):
+            raise ValueError("Request argument not found for handler")
         return req
 
     def handle_error(
@@ -166,7 +167,7 @@ class AIOHTTPParser(AsyncParser):
             content_type="application/json",
         )
 
-    def handle_invalid_json_error(
+    def _handle_invalid_json_error(
         self,
         error: typing.Union[json.JSONDecodeError, UnicodeDecodeError],
         req: Request,
--- a/src/webargs/asyncparser.py
+++ b/src/webargs/asyncparser.py
@@ -8,7 +8,6 @@ from collections.abc import Mapping
 from marshmallow import Schema, ValidationError
 from marshmallow.fields import Field
 import marshmallow as ma
-from marshmallow.utils import missing
 
 from webargs import core
 
@@ -22,53 +21,12 @@ class AsyncParser(core.Parser):
     either coroutines or regular methods.
     """
 
-    async def _parse_request(
-        self, schema: Schema, req: Request, locations: typing.Iterable
-    ) -> typing.Union[dict, list]:
-        if schema.many:
-            assert (
-                "json" in locations
-            ), "schema.many=True is only supported for JSON location"
-            # The ad hoc Nested field is more like a workaround or a helper,
-            # and it servers its purpose fine. However, if somebody has a desire
-            # to re-design the support of bulk-type arguments, go ahead.
-            parsed = await self.parse_arg(
-                name="json",
-                field=ma.fields.Nested(schema, many=True),
-                req=req,
-                locations=locations,
-            )
-            if parsed is missing:
-                parsed = []
-        else:
-            argdict = schema.fields
-            parsed = {}
-            for argname, field_obj in argdict.items():
-                if core.MARSHMALLOW_VERSION_INFO[0] < 3:
-                    parsed_value = await self.parse_arg(
-                        argname, field_obj, req, locations
-                    )
-                    # If load_from is specified on the field, try to parse from that key
-                    if parsed_value is missing and field_obj.load_from:
-                        parsed_value = await self.parse_arg(
-                            field_obj.load_from, field_obj, req, locations
-                        )
-                        argname = field_obj.load_from
-                else:
-                    argname = field_obj.data_key or argname
-                    parsed_value = await self.parse_arg(
-                        argname, field_obj, req, locations
-                    )
-                if parsed_value is not missing:
-                    parsed[argname] = parsed_value
-        return parsed
-
     # TODO: Lots of duplication from core.Parser here. Rethink.
     async def parse(
         self,
         argmap: ArgMap,
         req: Request = None,
-        locations: typing.Iterable = None,
+        location: str = None,
         validate: Validate = None,
         error_status_code: typing.Union[int, None] = None,
         error_headers: typing.Union[typing.Mapping[str, str], None] = None,
@@ -77,17 +35,18 @@ class AsyncParser(core.Parser):
 
         Receives the same arguments as `webargs.core.Parser.parse`.
         """
-        self.clear_cache()  # in case someone used `parse_*()`
+        self.clear_cache()  # in case someone used `location_load_*()`
         req = req if req is not None else self.get_default_request()
-        assert req is not None, "Must pass req object"
+        if req is None:
+            raise ValueError("Must pass req object")
         data = None
         validators = core._ensure_list_of_callables(validate)
         schema = self._get_schema(argmap, req)
         try:
-            parsed = await self._parse_request(
-                schema=schema, req=req, locations=locations or self.locations
+            location_data = await self._load_location_data(
+                schema=schema, req=req, location=location or self.location
             )
-            result = schema.load(parsed)
+            result = schema.load(location_data)
             data = result.data if core.MARSHMALLOW_VERSION_INFO[0] < 3 else result
             self._validate_arguments(data, validators)
         except ma.exceptions.ValidationError as error:
@@ -130,7 +89,7 @@ class AsyncParser(core.Parser):
         self,
         argmap: ArgMap,
         req: typing.Optional[Request] = None,
-        locations: typing.Iterable = None,
+        location: str = None,
         as_kwargs: bool = False,
         validate: Validate = None,
         error_status_code: typing.Optional[int] = None,
@@ -140,7 +99,7 @@ class AsyncParser(core.Parser):
 
         Receives the same arguments as `webargs.core.Parser.use_args`.
         """
-        locations = locations or self.locations
+        location = location or self.location
         request_obj = req
         # Optimization: If argmap is passed as a dictionary, we only need
         # to generate a Schema once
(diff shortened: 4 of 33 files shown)


CHANGELOG.rst
```diff
@@ -13,6 +13,56 @@ Refactoring:
...
+.. code-block:: python
+
+    # under webargs v5
+    class CompoundSchema:
+        q1 = ma.fields.Int(location="query")
+        q2 = ma.fields.Int(location="query")
+        h1 = ma.fields.Int(location="headers")
+
+
+    @parser.use_args(CompoundSchema(), locations=("query", "headers"))
```
sloria: Maybe use dicts instead of schema classes for these examples; I think that's the more common usage.

CHANGELOG.rst
```diff
@@ -13,6 +13,56 @@ Refactoring:
 * Don't mutate ``globals()`` in ``webargs.fields`` (:pr:`411`).
 * Use marshmallow 3's ``Schema.from_dict`` if available (:pr:`415`).
 
+* *Backwards-incompatible*: Schema fields may not specify a location any
+  longer, and `Parser.use_args` and `Parser.use_kwargs` now accept `location`
+  (singular) instead of `locations` plural. Instead of using a single field or
```
sloria:
```suggestion
  (singular) instead of `locations` (plural). Instead of using a single field or
```

src/webargs/asyncparser.py
```diff
@@ -22,53 +21,12 @@ class AsyncParser(core.Parser):
     either coroutines or regular methods.
     """
 
-    async def _parse_request(
```
sloria, line 25: This will be nice cleanup. 👍

src/webargs/flaskparser.py
```diff
@@ -48,17 +49,20 @@ class FlaskParser(core.Parser):
...
 
-    def parse_view_args(self, req, name, field):
-        """Pull a value from the request's ``view_args``."""
-        return core.get_value(req.view_args, name, field)
+    def location_load_view_args(self, req, schema):
+        """Pull a the request's ``view_args``."""
+        return req.view_args or core.missing
 
-    def parse_json(self, req, name, field):
+    def location_load_json(self, req, schema):
```
sloria: How about `get_json` etc.? It's less verbose and no less clear, IMO.

src/webargs/multidictproxy.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def _list_multiple_keys(self, schema):
+        result = []
+        for name, field in schema.fields.items():
+            if not is_multiple(field):
+                continue
+            if MARSHMALLOW_VERSION_INFO[0] < 3:
+                result.append(field.load_from or name)
+            else:
+                result.append(field.data_key or name)
```
sloria: `data_key` can be an empty string

```suggestion
                result.append(field.data_key if field.data_key is not None else name)
```

Same with `load_from`.

Just fixed this in marshmallow (https://github.com/marshmallow-code/marshmallow/pull/1379), so it's fresh in mind 😄

src/webargs/multidictproxy.py
```diff
@@ -0,0 +1,67 @@
...
+
+    In all other cases, __getitem__ proxies directly to the input multidict.
+    """
+
+    def __init__(self, multidict, schema):
+        self.data = multidict
+        self.multiple_keys = self._list_multiple_keys(schema)
+
+    def _list_multiple_keys(self, schema):
+        result = []
```
sloria: You can make this a `set` to optimize the membership checks.

src/webargs/testing.py
```diff
@@ -186,19 +201,29 @@ def test_parse_files(self, testapp):
...
         res = testapp.post(
-            "/echo",
+            "/echo_json",
             "",
             headers={"Accept": "application/json", "Content-Type": "application/json"},
+            expect_errors=True,
         )
-        assert res.status_code == 200
-        assert res.json == {"name": "World"}
+        assert res.status_code == 422
```
sloria: Agreed: this is probably the better behavior.

tests/test_core.py
```diff
@@ -254,42 +243,37 @@ def test_handle_error_reraises_errors(web_request):
...
         pass
 
     def error_handler(error, req, schema, status_code, headers):
         assert isinstance(schema, Schema)
         raise CustomError(error)
 
-    parse_json.side_effect = ValidationError("parse_json failed")
+    mock_schema = mock.Mock(spec=Schema)
+    mock_schema.strict = True
+    mock_schema.load.side_effect = ValidationError("parsing json failed")
```
sloria: Is the mocking here necessary?

sloria (commented):
Just drive-by's for now. Good work! Thanks for being so thorough with the docs.

I'll give this a proper lookover when I have more time to review.

src/webargs/flaskparser.py
sirosen: Yeah, I'm fine with that. But I think I want to keep the name `_get_loader` for the wrapper which looks up these methods in the `__location_map__`. I can't think of a better name for that.

Another option is `load_json` if you like that. I'll rename to `get_{location}` throughout and I don't mind doing another rename if you want.

src/webargs/multidictproxy.py
sirosen: I see the truthiness of your words. (See what I did there? :wink: )
I think this is the only place now that inspects individual fields on the schemas, but I'll double check to make sure this isn't an issue elsewhere.

src/webargs/multidictproxy.py
```diff
@@ -0,0 +1,67 @@
...
+        for name, field in schema.fields.items():
+            if not is_multiple(field):
+                continue
+            if MARSHMALLOW_VERSION_INFO[0] < 3:
+                result.append(field.load_from or name)
```
sirosen:
```suggestion
                result.append(field.load_from if field.load_from is not None else name)
```

Adding a self-suggestion so that I can batch this with the GitHub suggestion commit for the change below.

src/webargs/asyncparser.py
sirosen: Aw, shucks. Thanks. 😊 

The refactor of the core parser is less obvious ("are we just shuffling around complexity?") but when you see how much this one changes and how much the diff is negative, it builds my confidence that this is really simplifying things.

CHANGELOG.rst
sirosen: Sure, I see now that other parts of the changelog use dicts. I'll make it consistent.

src/webargs/flaskparser.py
sirosen: Actually, as I was in the midst of doing this, I saw that there are other functions like `get_default_request` and `get_request_from_view_args` which also have `get_` as a prefix. I'm going to go with `load_*` for now and we can discuss changing it. I'd like for the data-reading functions to have a distinct shared prefix that makes them identifiable.

tests/test_core.py
sirosen: This is a good catch. I'd rather use a real schema object where possible. Will fix.

src/webargs/flaskparser.py
sloria: Fine with me

src/webargs/flaskparser.py
```diff
@@ -72,31 +79,36 @@ def parse_json(self, req, name, field):
...
+        return json_data
+
+    def load_querystring(self, req, schema):
+        """Read query params from the request.
+
+        Is a multidict."""
+        return MultiDictProxy(req.args, schema)
+
+    def load_form(self, req, schema):
+        """Read form values from the request.
```
sloria: Maybe simplify to

```suggestion
        """Return form values from the request as a MultiDictProxy."""
```

tests/test_core.py
```diff
@@ -59,66 +59,70 @@ def parser():
...
-    assert parse_headers.call_count == 0
-    p.parse_arg("foo", field, web_request, locations=("headers",))
-    parse_headers.assert_called()
 
+@pytest.mark.skipif(
+    MARSHMALLOW_VERSION_INFO[0] < 3, reason="unknown=EXCLUDE added in marshmallow3"
+)
+def test_parse_with_excluding_schema(parser, web_request):
+    """
+    This is new in webargs 6.x ; it's the way you can "get back" the behavior
```
sloria: Change this to use code comments instead of a docstring so that pytest doesn't output this text.

src/webargs/asyncparser.py
```diff
@@ -77,17 +35,18 @@ class AsyncParser(core.Parser):
 
         Receives the same arguments as `webargs.core.Parser.parse`.
         """
-        self.clear_cache()  # in case someone used `parse_*()`
+        self.clear_cache()  # in case someone used `location_load_*()`
```
sloria, line 38:
```suggestion
        self.clear_cache()  # in case someone used `load_*()`
```

src/webargs/core.py
```diff
@@ -335,18 +238,18 @@ def parse(
 
          :return: A dictionary of parsed arguments
         """
-        self.clear_cache()  # in case someone used `parse_*()`
```
sloria, line 345: Why is this removed? I think it's still necessary.

src/webargs/multidictproxy.py
```diff
@@ -0,0 +1,67 @@
...
+            if val is None:
+                return None
+            else:
+                return [val]
+        return val
+
+    def __delitem__(self, key):
+        del self.data[key]
+
+    def __setitem__(self, key, value):
```
sloria, line 47: Should we make these immutable?  Flask's req.form, req.args, etc return [ImmutableMultiDicts](https://github.com/pallets/werkzeug/blob/6817cd093ee642192604fff163fed1be0e0a191e/src/werkzeug/datastructures.py#L1579) anyway.

tests/apps/flask_app.py
```diff
@@ -3,77 +3,100 @@
...
 import marshmallow as ma
-from webargs import fields
+from webargs import fields, dict2schema
 from webargs.flaskparser import parser, use_args, use_kwargs
 from webargs.core import MARSHMALLOW_VERSION_INFO
 
+if MARSHMALLOW_VERSION_INFO[0] < 3:
+    schema_kwargs = {"strict": True}
+else:
+    schema_kwargs = {"unknown": ma.EXCLUDE}
```
sloria: Is this necessary? Seems like the tests shouldn't be sending any unexpected keys.

tests/test_flaskparser.py
```diff
@@ -33,23 +33,31 @@ def test_use_args_with_view_args_parsing(self, testapp):
...
 
     # regression test for https://github.com/marshmallow-code/webargs/issues/145
     def test_nested_many_with_data_key(self, testapp):
-        res = testapp.post_json("/echo_nested_many_data_key", {"x_field": [{"id": 42}]})
-        # https://github.com/marshmallow-code/marshmallow/pull/714
+        post_with_raw_fieldname_args = (
+            "/echo_nested_many_data_key",
+            {"x_field": [{"id": 42}]},
+        )
+        # under marhsmallow2 this is allowed and works
```
sloria:
```suggestion
        # under marshmallow 2 this is allowed and works
```

sloria (commented):
Reviewed this a bit more. I'm basically good with the approach here.

I cut 5.5.0, so you'll need to update with `dev`. Feel free to merge instead of rebase; I intend to squash and merge when this is all done, if you don't mind.

I also made a 5.x-line branch so that we can maintain support for 5.x for some time after we release 6.0.

src/webargs/flaskparser.py
sirosen: 👍 I like this wording. I'll go through and update them all when I next get time to work on this.

src/webargs/core.py
sirosen: That was a mistake which slipped through. I'll amend to revert this change (it's part of ab05a80c ).

In my very first drafts, I was playing with changing things about how the cache worked but it was just too much unrelated refactoring and noise. Apparently I accidentally added one of those changes.

sloria:
Should we rename `@location_handler` to `@location_loader` for consistency with the `load_*` naming?

Just a thought: doesn't need to happen in this PR.

src/webargs/multidictproxy.py
sirosen: I don't think so, because it will fail with an appropriate error the way that this is now. Replicating the work in the frameworks to make these immutable just seems to offer little value to me.

If a Flask user ended up handling the `MultiDictProxy` (e.g. a validator which takes original data), using `__setitem__` will give a `TypeError` because it just calls `ImmutableMultiDict.__setitem__` anyway. That seems like pretty good behavior to me.

Is there a case in which this will fall over?

I am looking at those classes and wondering if we should proxy the `copy` method for Flask though. I wouldn't be surprised if other frameworks have useful specialty-methods which people might want to use too.

tests/apps/flask_app.py
sirosen: There's currently a test which explicitly checks that an unexpected key is ignored -- behavior which shouldn't be the norm going forward. So for now this is just duct-tape to make that test work.

I think we want tests, probably with `skipif`s for marshmallow 2, which explores the various `unknown` settings you could pass to the schema, to replace this.

I'll make that change a priority, so that we can talk about whether or not we like those tests.

src/webargs/multidictproxy.py
sloria: Fair point; I think this is fine as is. I'm not sure about `copy`; I don't know when that would come up in practice.

sirosen: Maybe the best option for `copy` and its ilk is to ignore it. And if someone comes along requesting that `MultiDictProxy` also proxies `copy`, it can be added pretty trivially.

sloria: Sounds good

tests/test_flaskparser.py
sirosen: This was done in 176a9f7 (and I marked co-authorship the way github does out of curiosity, to see if it would understand and resolve this suggestion).
It doesn't seem to have, so I'm just going to mark the accepted suggestions as resolved.

src/webargs/core.py
sirosen: Fixed this in recent changes ; thanks for spotting it.

(I'd rebase, but if you plan to squash, it's easier to review in new commits).

sirosen:
> Should we rename `@location_handler` to `@location_loader` for consistency with the `load_*` naming?

I don't mind changing it or keeping it the same. There isn't a huge difference between these names.

One advantage of renaming is that it would immediately fail for webargs v5 users who are using `@location_handler` today and upgrade. That might be reason enough to make the change.

tests/apps/flask_app.py
sirosen: 05d308e changes this around a bit.
The `unknown=EXCLUDE` is now used (1) in an explicit "ignores extra" route and (2) in `/echo_headers`, to handle the presence of various headers not explicitly set by the client.

sloria:
Up to you. Feel free to do the `location_loader` rename, or we can do it in a follow-up PR.

sirosen:
> Up to you. Feel free to do the `location_loader` rename, or we can do it in a follow-up PR.

I'll probably do it in here, since it's already a large, breaky change.

---

Something that I've started to realize poses an issue: the way that I've been writing things, such that webargs treats an empty JSON payload as `missing`, actually makes the behavior more confusing and non-uniform.

The issue is that `json` is the *only* location which can behave this way. No query params or an empty form post needs to be interpreted as `{}`.

I'm working on a refactor now in which `load_json` will still return `missing`, but `_load_location_data` (the wrapper which gets the loader func and calls it) will check if the result is `missing` and return `{}` if it is.
This lets webargs can still pass around information about the presence or absence of data, but when data is being handed off to `Schema.load` it needs to be "normalized" for consumption.

Part of the goal is to ensure it's easy to add something like
```python
def load_json_or_form(self, req, schema):
    data = self.load_json(req, schema)
    if data is missing:
        data = self.load_form(req, schema)
    return data
```

and I think this achieves that.

This will have a public-facing impact on how webargs v6 behaves, but I think generally for the better. It will make it easier to use webargs in APIs with optional request bodies.
So expect it to come with numerous changes to tests' expectations around posting empty data to `/echo_json` endpoints.

sirosen:
Er, to clarify, I think the above is important to be very clear and open about because it constitutes another reversal on #297 (back to the current behavior)

The first version of this PR changed the decision on that saying that `""` is invalid JSON.
I thought this change would be necessary for implementing `json_or_form`, and it kind of appeals as more pedantically correct. But as I work on things I realize that this has negative impact on webargs' usability, and it's not as necessary as I thought if we distinguish between the value returned by `load_json` and the value seen by `Schema.load`.

One of the big cases that changes my mind is the idea of an endpoint with an optional JSON body.
There are a number of behaviors currently accepted for a missing request body, and it's not really obvious what should be allowed on this front.

sirosen:
As of f005c40 , this is finally be passing tests, as all of the parser conversions are finished. And I've just merged `dev` back in again to get this up to date.

This is what I'd consider the very minimum to be safe to merge. But I still plan to do two more things before removing `WIP` from the PR title:
- Update narrative docs with new usage
- Add some explicit testing of various `unknown=...` behaviors

src/webargs/aiohttpparser.py
```diff
@@ -147,7 +148,7 @@ def handle_error(
         schema: Schema,
         error_status_code: typing.Union[int, None] = None,
         error_headers: typing.Union[typing.Mapping[str, str], None] = None,
-    ) -> "typing.NoReturn":
```
sloria: This is actually necessary to account for older patch versions of Python 3.5 https://github.com/marshmallow-code/webargs/issues/366

sloria (commented):
Excellent. Let me know when you're confident in merging this.

src/webargs/aiohttpparser.py
sirosen: That makes me sad, but makes sense. I'll switch it back.

sloria: Yeah.. I'm getting kinda exhausted by [these](https://github.com/marshmallow-code/marshmallow/pull/1410) =/

sloria: Alternatively, we could conditionally install `typing` on Python 3.5. Something like:

```python
'typing>=3.7.4; python_version<"3.5.4"',
```

I think I'm going to end up doing this in marshmallow. 

Doesn't have to happen in this PR though.

sloria: Disregard that; adding the typing backport doesn't fix it. It's only useful for Python versions less than 3.5.0.

sirosen:
I've just removed `[WIP]` from the title because I think this is ready.
I've made changes to the narrative docs (more content was removed than added; I am sorry), and added some more tests.

As an important final change, I was trying to add `json_or_form` as a location -- to validate that the approach I have for that will work -- and my testing revealed some non-uniform treatment of empty JSON payloads between different frameworks.

I've added https://github.com/marshmallow-code/webargs/pull/420/commits/b8d3fee2c9d332c3ca3f3ba49ea0c483da18a590 , which makes the decision to check content-type on the frameworks where this was not done (django and pyramid) and return `missing` if it is non-JSON.

docs/quickstart.rst
```diff
@@ -243,7 +237,7 @@ Nesting Fields
 
 .. note::
 
-    By default, webargs only parses nested fields using the ``json`` request location. You can, however, :ref:`implement your own parser <custom-parsers>` to add nested field functionality to the other locations.
+    Of the default supported locations in webargs, only the ``json`` request location supports nested datastrctures. You can, however, :ref:`implement your own data loader <custom-loaders>` to add nested field functionality to the other locations.
```
lafrech: s/datastrctures/datastructures/

sloria:
Great. I probably won't be able to review this until this weekend at the earliest. Thanks for all your work on this.

docs/quickstart.rst
sirosen: Good catch. Let me patch this right now so I don't forget.

sirosen:
Just as a short FYI: I'm going to be taking a couple of weeks off, and likely won't be able to respond to any feedback for the next two weeks or so anyway. So please don't worry if you write some comments and don't see a reply -- this isn't abandoned, and I'm still invested in seeing it through.

sirosen:
In light of the fact that it motivates b8d3fee , I've [added the `json_or_form` location to this branch](https://github.com/marshmallow-code/webargs/pull/420/commits/1d352d2ad49e1eb503db97e32f88f73eebe16de7). The former commit doesn't make as much sense without the latter.

lafrech:
Thank you so much @sirosen for tackling this and sorry for the so long delay.

I've tried to port flask-smorest to your webargs branch and I only had two adaptations to make:

- replace `locations=[location]` with `location=location` in a call to use_args
- add `location='query'` to a call to `parser.parse`

Those are known breaking changes. My point is that I didn't discover any issue.

Then I tried to port the application we're working on. I had to

- fix `NestedQueryArgsParser`. We copied it from webargs docs and since you also fixed the docs we just had to copy your new version. Great.
- add missing `content_type='application/json'` in some test functions

Again, nothing surprising.

I shall now actually review the code, but at least it proved functional on our real-life cases.

Thanks again.

sirosen:
I'll take some time soon (hopefully next few days) to revisit the changelog and make sure it's accurate. In particular

> add missing `content_type='application/json'` in some test functions

made me realize that two parsers have changed (the `DjangoParser` and `PyramidParser`) to require `Content-Type` where they did not before. That's currently missing from the changelog.

---

Anyway, please don't worry about taking time with this! I _more than_ understand how hard it can be to get time to work on open source projects and properly review big PRs. 🙂

lafrech:
Thanks.

I'm using Flask, BTW, so I guess this also changed on  `FlaskParser`.

lafrech:
I could be wrong but here's my understanding:

- In the Flask client, if I forget `content_type='application/json'`, the payload is sent as form data.
- webargs used to check several locations by default and now it only checks `json`.

So the breaking change would only be the default location checked for input.

sirosen:
I just amended the changelog for 6.0 based on a reread of the changes. Listed as a bugfix (because it makes webargs internally consistent), several parsers now require `Content-Type` to be set for JSON payloads. Since anything matching `application/(.*\+)?json` will do, I didn't specify exactly what it must be set to.

`FlaskParser` was one of the parsers with changes. Even if you send it JSON data, if you don't sent the `Content-Type` header it will not be parsed as JSON.
That jives with what you're saying you observed, so I think all is well.

---

There are new changes in `dev`, but they didn't merge back into this branch smoothly (quickly "fixing" the conflicts produced broken tests). I'll have to look more closely at the changes to understand what went wrong.

sirosen:
I've just rebased and pushed a cleaned-up version of this branch which works against `dev`. (I tried doing this merge-back thing, but it's just not how I'm used to using git. The commit history becomes harder to navigate.)

My attempt to handle this by merging things yesterday failed because I did not properly incorporate #428 . Because this PR introduces the idea of a shared `load_json` handler in the core parser, that needed to be updated as well.

lafrech:
Thanks.

Rebase is better. Sometimes it is harder to achieve (too many changes and too many commits in the branch to rebase), so a merge is acceptable. If rebase is easier for you, then that's just great.

I'd really like to merge this. I'll try to review this as soon as I can. Can't promise when.

docs/quickstart.rst
```diff
@@ -23,17 +23,11 @@ Arguments are specified as a dictionary of name -> :class:`Field <marshmallow.fi
...
         # or you want to rename a key
-        "content_type": fields.Str(load_from="Content-Type", location="headers"),
+        "user_type": fields.Str(load_from="user-type"),
         # OR, on marshmallow 3
-        # "content_type": fields.Str(data_key="Content-Type", location="headers"),
-        # File uploads
-        "profile_image": fields.Field(
-            location="files", validate=lambda f: f.mimetype in ["image/jpeg", "image/png"]
-        ),
+        # "content_type": fields.Str(data_key="user-type"),
```
lafrech: s/content_type/user_type/ ?

src/webargs/core.py
```diff
@@ -531,41 +446,98 @@ def handle_error(error, req, schema, status_code, headers):
...
+            except json.JSONDecodeError as e:
+                if e.doc == "":
+                    return missing
+                else:
+                    return self._handle_invalid_json_error(e, req)
+            except UnicodeDecodeError as e:
+                return self._handle_invalid_json_error(e, req)
+
+        return json_data
+
```
lafrech, line 482: Why not also cache `missing`?

```py
        json_data = self._cache.get("json")
        if json_data is None:
            try:
                json_data = self._raw_load_json(req)
            except json.JSONDecodeError as e:
                if e.doc == "":
                    json_data = missing
                else:
                    return self._handle_invalid_json_error(e, req)
            except UnicodeDecodeError as e:
                return self._handle_invalid_json_error(e, req)

        self._cache["json"] = json_data
        return json_data
```

src/webargs/core.py
```diff
@@ -329,9 +239,9 @@ def parse(
             of argname -> `marshmallow.fields.Field` pairs, or a callable
             which accepts a request and returns a `marshmallow.Schema`.
         :param req: The request object to parse.
-        :param tuple locations: Where on the request to search for values.
-            Can include one or more of ``('json', 'querystring', 'form',
-            'headers', 'cookies', 'files')``.
+        :param str location: Where on the request to load values.
+            Can be one of ``('json', 'querystring', 'form', 'headers', 'cookies',
+            'files')``.
         :param callable validate: Validation function or list of validation functions
```
lafrech, line 246: The list is incomplete. Should we point to `__location_map__` instead?

src/webargs/multidictproxy.py
```diff
@@ -0,0 +1,67 @@
...
+            elif hasattr(self.data, "getall"):
+                return self.data.getall(key)
+            elif isinstance(val, (list, tuple)):
+                return val
+            if val is None:
+                return None
+            else:
+                return [val]
+        return val
+
```
lafrech, line 43: no-else-return: Unncessary "else" after "return"

src/webargs/multidictproxy.py
```diff
@@ -0,0 +1,67 @@
...
+from webargs.core import missing, is_multiple
+
+
+class MultiDictProxy(Mapping):
+    """
+    A proxy object which wraps multidict types along with a matching schema
+    Whenever a value is looked up, it is checked against the schema to see if
+    there is a matching field where `is_multiple` is True. If there is, then
+    the data should be loaded as a list or tuple.
+
```
lafrech, line 11: I would have been tempted to subclass rather than proxying, to only override what is needed, but I suppose there's a good reason for proxying.

src/webargs/falconparser.py
```diff
@@ -91,12 +75,20 @@ def to_dict(self, *args, **kwargs):
...
-        """Pull a querystring value from the request."""
-        return core.get_value(req.params, name, field)
+    # Note on the use of MultiDictProxy throughout:
+    # Falcon parses query strings and form values into ordinary dicts, but with
+    # the values listified where appropriate
+    # it is still therefore necessary in these cases to wrap them in
+    # MultiDictProxy because we need to use the schema to determine when single
+    # values should be wrapped in lists due to the type of the destination
+    # field
+
```
lafrech, line 85: I wasn't sure about this so I removed the `MultiDictPRoxy`. No test failed.

I didn't investigate any further but maybe a test could be added.

lafrech (commented):
Looks good to me.

I made a few comments.

I didn't review the async parsers.

I lean towards releasing as 6.0 rather than as a 6.0b pre-release.

docs/quickstart.rst
sirosen: I need to take another look at this, but obviously there's a typo here.

I think really that I'd like to revisit the quickstart after working on stacked `use_args` errors ([per this comment from sloria](https://github.com/marshmallow-code/webargs/issues/419#issuecomment-525765453)), since that will make it possible to have use-cases in here with
```python
@use_args(...)
@use_args(...)
def view_func(...):
    ...
```

(Those work today, but error semantics are based on the order of the decorators.)

src/webargs/falconparser.py
sirosen: It's hard to remember perfectly, but based on the comment I think this is something to do with the interplay between `marshmallow` Lists and Falcon parsing.
I need to go in and add a test to try to capture this.

src/webargs/multidictproxy.py
sirosen: It's hard to be sure exactly what I was thinking when I wrote the first version, but I think that proxying ends up being less code. If you subclass the multidict types for Flask, Django, etc. you end up with a variety of classes to handle.

Also, if I'm right that Falcon really does need MultiDictProxy for List fields (I need to take some time to prove this out), then we basically end up writing this up anyway, since Falcon doesn't have any native multidict type.

src/webargs/multidictproxy.py
sirosen: I'm not sure I follow. You mean it should read
```python
if val is None:
    return None
return [val]
```
?

I don't mind doing that if you think it reads cleaner to drop the `else` (or if its a general style rule for certain style guides).

sirosen (commented):
I wanted to get to the review feedback this past weekend, but it didn't happen. I'll try to find a good time for it soon. There's one item where I had a question to make sure I correctly understand what's wanted.

src/webargs/multidictproxy.py
lafrech: This is something that pylint warns about. Removing the else after every return makes it clearer that the return line exits the function. At least it is the rationale.

```py
    def __getitem__(self, key):
        val = self.data.get(key, missing)
        if val is not missing and key in self.multiple_keys:
            if hasattr(self.data, "getlist"):
                return self.data.getlist(key)
            if hasattr(self.data, "getall"):
                return self.data.getall(key)
            if isinstance(val, (list, tuple)):
                return val
            if val is None:
                return None
            return [val]
        return val
```

or even

```py
    def __getitem__(self, key):
        val = self.data.get(key, missing)
        if val is missing or key not in self.multiple_keys:
            return val
        if hasattr(self.data, "getlist"):
            return self.data.getlist(key)
        if hasattr(self.data, "getall"):
            return self.data.getall(key)
        if isinstance(val, (list, tuple)):
            return val
        if val is None:
            return None
        return [val]
```

docs/quickstart.rst
sirosen: I reread the doc. It turns out, fixing to just read `user_type`, as you suggested, is simplest.

src/webargs/multidictproxy.py
sirosen: I like the second version better because it's less nested, so I switched to that.

I tried to take a look at what pylint has to say to see if there's anything else I can trivially clean up, but I pretty much instantly re-learned why I find pylint so annoying. 😂 
(It's a good tool and all, but it is _picky_!)

src/webargs/falconparser.py
sirosen: I confirmed that there's a failing case if you remove this and I've added a test for it.
If you remove `MultiDictProxy` and pass a _single value_ to a `List` field, then things will fail.

The reason is that Falcon parses query params like so:
```
/path?foo=1&foo=2
  =>  {"foo": [1, 2]}
/path?foo=1
  =>  {"foo": 1}
```
When we try to convert `{"foo": 1}` to a list field, we run aground on `1` as a non-list value.

sirosen (commented):
I just pushed a commit with fixes for the several issues found in the last review.
I'm going to do a final double-check on everything and (hopefully) mark all of the outstanding comments as resolved.

src/webargs/core.py
sirosen: I missed this in my pass over things today.
If I had to guess, I'd say I was just imitating what came before. I see no reason not to cache `missing` results -- I'll make the change.

src/webargs/core.py
sirosen: I'm going to take a look at this, but my instinct is to both write it all out here and point there in addition... In terms of the documentation's readability, I think it helps to have the list of available values included right there.

src/webargs/core.py
sirosen: I've just changed this. One test which explicitly checked `self._cache` needed to be tweaked, but nothing else broke with the change.
It's simpler, but I noted in the commit message that we're changing it in part because we don't have context here for why we _shouldn't_ cache `missing`. So if it turns out, on review, that there's some issue which comes up in webargs 6.x because of it, we won't be worried about reverting the change.

src/webargs/core.py
sirosen: I've made a change to do both. I was a little concerned that there might be other locations that need to be kept in sync, but it looks like this is the only place where the contents of `__location_map__` are spelled out.

I think this is an okay solution, so I'm going to mark this resolved. But I don't mind discussing further if you like -- e.g. if you're worried that it will get out of sync again in the future.

sirosen:
Okay, I missed a couple of small things, so I handled those in follow-up commits.

I'm excited about merging this in part because I want to get cracking on the changes to `@use_args` I pitched in #419 for us to review. 😄

lafrech:
Great.

@sloria, should we release a beta to give frameworks time to adapt to the changes?

Then we'd release a final 6.0.0 with [Python 2 dropped](#440).

sloria:
Haven't had time to review the latest changes, so I'll defer to @lafrech on merging this. Releasing a 6.0.0b1 makes sense to me

lafrech:
Alright. Merging. I'll try to release a beta version today.

lafrech:
@sloria, I just got this error again.

```
remote: Resolving deltas: 100% (4/4), completed with 4 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/dev.
remote: error: Required status check "marshmallow-code.webargs" is expected.
To ssh://github.com/marshmallow-code/webargs
 ! [remote rejected] dev -> dev (protected branch hook declined)
error: failed to push some refs to 'ssh://git@github.com/marshmallow-code/webargs'
```

Am I missing privileges on this repo?

sloria:
@lafrech I just gave you admin privs. Can you try again?

PR #432, opened 2019-10-10
Add documentation about marshmallow-dataclass

CptSpaceToaster:
Webargs already works out of the box with [marshmallow-dataclass](https://github.com/lovasoa/marshmallow_dataclass), however there are some [tripping points](https://github.com/lovasoa/marshmallow_dataclass/issues/42) one can run into if you miss the warnings associated with `use_kwargs`, which expects all schemas to deserialize into a dictionary. An assumption not maintained by marshmallow-dataclass _by design_. A reminder to rely on `use_args` instead could help alleviate the confusion I experienced.

I've included a simple example of how I would use marshmallow-dataclass demonstrating how to deserialize requests into dataclasses with type hints.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -74,10 +74,13 @@ When you need more flexibility in defining input schemas, you can pass a marshma
     If you're using marshmallow 2, you should always set ``strict=True`` (either as a ``class Meta`` option or in the Schema's constructor) when passing a schema to webargs. This will ensure that the parser's error handler is invoked when expected.
 
 
-When not to use `use_kwargs`
-----------------------------
+When to avoid `use_kwargs`
+--------------------------
 
-Any Schema passed to use_kwargs MUST deserialize to a dictionary of data. If you are relying on schemas that provide `post_load <marshmallow.decorators.post_load>` methods that end up transforming the result, then you should instead rely on use_args.
+Any  `Schema <marshmallow.Schema>` passed to `use_kwargs <webargs.core.Parser.use_kwargs>` MUST deserialize to a dictionary of data.
+If your schema has a `post_load <marshmallow.decorators.post_load>` method 
+that returns a non-dictionary,
+you should use `use_args <webargs.core.Parser.use_args>` instead.
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


sloria (changes requested):
Perhaps the documentation should be more generic, since the issue is not really specific to marshmallow-dataclass. This is relevant when using any schema that implements a `post_load` method that returns an object rather than a dict.

CptSpaceToaster:
I'm totally up for making changes, however the issue is technically captured by the warning at the end of the [marshmallow integration](https://webargs.readthedocs.io/en/latest/advanced.html#marshmallow-integration) paragraph.

> **Warning**
> Any Schema passed to use_kwargs MUST deserialize to a dictionary of data. Keep this in mind when writing post_load methods.

For users writing their own post load methods, the documentation is _already_ trying to clear up those tripping points.  Maybe we could expand that warning into an entire section, and explicitly call out `marshmallow-dataclass` as an example?

CptSpaceToaster:
Maybe we could remove the warning, and write a section like this?

> When not to use `use_kwargs`
> -----------------------
> 
> Any Schema passed to use_kwargs MUST deserialize to a dictionary of data. If you are relying on schemas that provide `post_load` methods that end up transforming the result, then you should instead rely on use_args.
> 
> For example, marshmallow dataclass can be used to provide schemas built from python3 dataclasses. However, these schemas will deserialize content into a dataclass and NOT a dictionary.
> 
> ```python
>     from marshmallow_dataclass import dataclass
>     from webargs.flaskparser import use_args
> 
> 
>     @dataclass
>     class Greeting:
>         name: str
>         prefix: str = "Hello"
> 
> 
>     @use_args(Greeting.Schema)
>     def profile_posts(greeting: Greeting):
>         return f"{greeting.prefix} {greeting.name}"
> ```

`use_args` would link to the proper link in the doc. I'm tempted to make [marshmallow dataclass](https://github.com/lovasoa/marshmallow_dataclass) link to that project's github, but I'd like permission from maintainers here before I just assume that'd be Ok.

sloria:
@CptSpaceToaster I like that direction. Maybe we should exemplify usage with marshmallow-sqlalchemy? From what I can tell, marshmallow_dataclass isn't as popular, whereas I think many webargs users are already ma-sqla users. Perhaps you could write an example with a plain schema with a `post_load` method that returns an object, then include mention of libraries that generate schema (ma-sqla, ma_dataclass).

CptSpaceToaster:
Would love some feedback to see if the grammar and language are suitable!

sloria:
Thanks!

PR #446, opened 2019-12-14
Add tests / test files to sdist

koobs:
Add tests and related test files to the MANIFEST.in in order to include them in the PyPI sdist

This allows downstream users, including OS packagers to test the package using the source distribution

Note: setup.py:packages already includes an exclusion for packages matching test* to preclude it from being installed

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -41,3 +41,5 @@ Contributors (chronological)
 * Jonathan Angelo <https://github.com/jangelo>
 * @zhenhua32 <https://github.com/zhenhua32>
 * Martin Roy <https://github.com/lindycoder>
+* Kubilay Kocak <https://github.com/koobs>
+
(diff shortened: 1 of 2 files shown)


lafrech:
Would you need this done in [marshmallow](https://github.com/marshmallow-code/marshmallow/blob/dev/MANIFEST.in) too?

koobs:
@lafrech That would be great, and in any other projects you can contribute to. We (FreeBSD and other OS packagers) make this request, generically (include test/test related files in sdist), all the time :) 

Much appreciated

lafrech:
I have no objection to this.

@sloria, what do you think. Is this something you could do with all-repos on all the repos?

sloria:
Let's go ahead and merge this. I'll all-repos this later when I have some time

sloria:
On second thought, does this just copy `tests` as a top-level package in site-packages? If so, couldn't this mess up users' environments? See https://github.com/lovasoa/marshmallow_dataclass/pull/59 .

sloria:
Ah, nvm; confirmed that `tests` doesn't get included in site-packages--it's only in the distribution. 👍

Issue #448, opened 2019-12-17
falconparser is broken


dodumosu:
`falconparser` uses `falcon.util.uri.parse_query_string`, but the keyword argument is `keep_blank`, not `keep_blank_qs_values`

alessandromarchetti:
Hi, any news on when this change will be accepted?

lafrech:
Fixed in 6.0.0b3. (#459)

PR #449, opened 2019-12-17
fix: rename incorrect keyword argument to `parse_query_string`

dodumosu:
`falconparser` uses `falcon.util.uri.parse_query_string`, but uses
an incorrect keyword argument name, `keep_blank_qs_values`. the correct
argument name is `keep_blank`

see: #448

### DIFF
--- a/src/webargs/falconparser.py
+++ b/src/webargs/falconparser.py
@@ -13,6 +13,9 @@ HTTP_422 = "422 Unprocessable Entity"
 status_map = {422: HTTP_422}
 
 
+FALCON_VERSION = falcon.__version__[0]
+
+
 # Collect all exceptions from falcon.status_codes
 def _find_exceptions():
     for name in filter(lambda n: n.startswith("HTTP"), dir(falcon.status_codes)):
@@ -65,9 +68,15 @@ def parse_form_body(req):
                 "will be ignored."
             )
 
+        if FALCON_VERSION == "1":
+            key = "keep_blank_qs_values"
+        elif FALCON_VERSION == "2":
+            key = "keep_blank"
+        kwargs = {key: req.options.keep_blank_qs_values}
+
         if body:
             return parse_query_string(
-                body, keep_blank=req.options.keep_blank_qs_values
+                body, **kwargs
             )
     return {}
 
--- a/src/webargs/falconparser.py
+++ b/src/webargs/falconparser.py
@@ -15,7 +15,7 @@ HTTP_422 = "422 Unprocessable Entity"
 status_map = {422: HTTP_422}
 
 
-FALCON_VERSION = tuple(LooseVersion(falcon.__version__).version)
+FALCON_VERSION_INFO = tuple(LooseVersion(falcon.__version__).version)
 
 
 # Collect all exceptions from falcon.status_codes
--- a/src/webargs/falconparser.py
+++ b/src/webargs/falconparser.py
@@ -67,7 +67,7 @@ def parse_form_body(req):
 
         if body:
             return parse_query_string(
-                body, keep_blank_qs_values=req.options.keep_blank_qs_values
+                body, keep_blank=req.options.keep_blank_qs_values
             )
     return {}
 
--- a/src/webargs/falconparser.py
+++ b/src/webargs/falconparser.py
@@ -70,13 +70,12 @@ def parse_form_body(req):
                 "will be ignored."
             )
 
-        if FALCON_VERSION_INFO[0] < 2:
-            key = "keep_blank_qs_values"
-        else:
-            key = "keep_blank"
-        kwargs = {key: req.options.keep_blank_qs_values}
-
         if body:
+            if FALCON_VERSION_INFO[0] < 2:
+                key = "keep_blank_qs_values"
+            else:
+                key = "keep_blank"
+            kwargs = {key: req.options.keep_blank_qs_values}
             return parse_query_string(body, **kwargs)
     return {}
 
--- a/src/webargs/falconparser.py
+++ b/src/webargs/falconparser.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """Falcon request argument parsing module.
 """
+from distutils.version import LooseVersion
+
 import falcon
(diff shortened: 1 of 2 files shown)


src/webargs/falconparser.py
```diff
@@ -13,6 +13,9 @@
 status_map = {422: HTTP_422}
 
 
+FALCON_VERSION = falcon.__version__[0]
+
```
lafrech: We could copy compat.py and do

FALCON_VERSION_INFO = tuple(LooseVersion(falcon.__version__).version)

but I don't think there is much interest in doing so.

Let's keep it as is.

src/webargs/falconparser.py
```diff
@@ -13,6 +13,9 @@
...
+
```
lafrech, line 20: On second thought, I'd write

    FALCON_VERSION_INFO = tuple(LooseVersion(falcon.__version__).version)

so that we get version as integer.

(We could also cast to `int`. I don't really mind.)

src/webargs/falconparser.py
```diff
@@ -65,9 +68,15 @@ def parse_form_body(req):
                 "will be ignored."
             )
 
+        if FALCON_VERSION == "1":
+            key = "keep_blank_qs_values"
+        elif FALCON_VERSION == "2":
+            key = "keep_blank"
+        kwargs = {key: req.options.keep_blank_qs_values}
```
lafrech: I'd write

```py
        if FALCON_VERSION_INFO < 2:
            key = "keep_blank_qs_values"
        else:
            key = "keep_blank"
```

This way, it won't break in Falcon 3.

src/webargs/falconparser.py
```diff
@@ -65,9 +68,15 @@ def parse_form_body(req):
...
+
         if body:
             return parse_query_string(
-                body, keep_blank_qs_values=req.options.keep_blank_qs_values
+                body, **kwargs
             )
     return {}
```
lafrech, line 80: A single line should be enough.

lafrech (changes requested):
Good.

I think it makes sense to maintain compatibility with Falcon 1. Flacon 2 is really recent and backward compat is cheap.

Please also add yourself to AUTHORS.

src/webargs/falconparser.py
```diff
@@ -13,7 +15,7 @@
 status_map = {422: HTTP_422}
 
 
-FALCON_VERSION = falcon.__version__[0]
+FALCON_VERSION = tuple(LooseVersion(falcon.__version__).version)
```
lafrech: s/FALCON_VERSION/FALCON_VERSION_INFO/

Or at least be consistent with below.

lafrech (changes requested):
Great. Just one typo to fix.

src/webargs/falconparser.py
```diff
@@ -65,10 +70,14 @@ def parse_form_body(req):
                 "will be ignored."
             )
 
+        if FALCON_VERSION_INFO[0] < 2:
+            key = "keep_blank_qs_values"
+        else:
+            key = "keep_blank"
+        kwargs = {key: req.options.keep_blank_qs_values}
```
lafrech: The whole `if FALCON_VERSION_INFO[0] < 2:` block could be moved under the `if body` condition.

```py
        if body:
            if FALCON_VERSION_INFO[0] < 2:
                key = "keep_blank_qs_values"
            else:
                key = "keep_blank"
            return parse_query_string(body, **{key: req.options.keep_blank_qs_values})
```

lafrech (changes requested):
One last change I didn't think of before.

Thanks.

lafrech:
If we mean to support both Flacon 1 and 2, we should ensure the tests test both version. @dodumosu if you're not comfortable with modifying CI tests config, we'll do it when we get the time.

Is Flacon 1 still maintained? Shall we drop support for it in the next webargs major version ?

Nateyo:
As far as I know, most of the development going on with falcon is now 2.0+. This PR should likely update the setup.py as well since right now it specifies `falcon>=1.4.0,<2.0`. There's actually another problem with args to `req.stream.read()` which I can submit a PR for, building on top of @dodumosu's changes

lafrech:
What about keeping webargs 5.x Falcon 1.x and dropping Falcon 1.x in webargs 6.0?

|webargs|Falcon|
|-|-|
|5|1 only|
|6|2 only|

lafrech:
@sloria, we could

- [x] merge #420
- [x] drop Python 2
- [ ] apply the change in this PR but just keep Flacon 2 compatibility, update setup.py (and CI config if needed)
- [ ] fix `req.stream.read()` as per [this comment](https://github.com/marshmallow-code/webargs/pull/449#issuecomment-569941269)

We could publish beta releases along the way.

sloria:
@lafrech I'm good with that plan

lafrech:
Guys, I just merged #420.

You may rebase this and drop Falcon 1 (or start from scratch if it is easier), and send another PR to fix `req.stream.read()`.

Thanks.

lafrech:
~Better wait until #454 is merged to avoid merge conflicts.~ Done.

lafrech:
@dodumosu @Nateyo are you willing to tackle this?

No pressure, just asking. I could rebase the change in this PR when I get the time, but I don't know what @Nateyo has in mind regarding `req.stream.read()` and I don't use Falcon so ideally I'd rather defer that to you.

Basically, change setup.py to require falcon>=2 and do all required changes (the two expressed above + others if needed).

Nateyo:
@lafrech Sure, I can work on it this week. I can cherry pick these commits and submit a whole new PR to the dev branch? Would that be easier since you want 5.x to keep Falcon 1 compat?

lafrech:
Great. Thanks @Nateyo.

Further dev on webargs 5 will take place in branch `5.x-line`. We're dropping Falcon 1 in `dev` branch. I wouldn't cherry-pick and just start from scratch as the changeset is much smaller when dropping Falcon 2 than when keeping compatibility. You may submit a single PR with the whole Falcon 2 move.

lafrech:
Superseded by #459.

Issue #440, opened 2019-11-13
Drop Python 2


lafrech:
Let's wait on #420 to be merged first to avoid a painful rebase.

- [x] Drop Python 2
- [x] Make arguments keyword-only where sensible

Do we also want to drop Python 3.5 ? This can be done in a later major release.

sloria:
Wrote up a checklist recently for this: https://gist.github.com/sloria/b4a5fc2281f86a5733af198d609c5ab3

sirosen:
I wanted to chime in with my $0.02: I wouldn't drop py3.5 support yet because it's the py3 version shipped on several distro versions, notably Ubuntu 16.04.
Since so many people are using Ubuntu as their first distro, and there's the WSL stuff on Windows, there's some reason to give the Ubuntu lifecycle "priority" when looking at the landscape of distros.

By way of example, if someone is using webargs but they're having trouble upgrading from Ubuntu 16.04 infrastructure to Ubuntu 18.04, then they wouldn't be able to move to the new webargs without either (A) fixing their platform version issues or (B) installing another python version alongside the system python. (Yes, they should fix their problems, but my experience is that by the time you're even _talking_ about being stuck on an old platform version, you're in trouble.)

The same argument doesn't apply to forcing the py2 -> py3 transition -- I'm 👍 on dropping py2.

lafrech:
Yeah, no pressure to remove py35 soon. Can be done later.

sloria:
Sounds good. I'm fine keeping support for 3.5 for now.

lafrech:
Summoning @hugovk again.

Would you be willing to tackle this?

Just asking. I don't know if you even use webargs.

hugovk:
Happy to help with Python 3 migration!

Please see PR #454 for the first checkbox item: " Drop Python 2".

The second checkbox "Make arguments keyword-only where sensible" can follow in another PR.

sirosen:
With #472 merged, we can call this done!

I just did a check on the repo looking for _anything_ python2-related I could find and it came up blank.

PR #454, opened 2020-01-03
Drop support for EOL Python 2.7

hugovk:
For https://github.com/marshmallow-code/webargs/issues/440.


lafrech:
Thank you so much for your work on this, @hugovk.

hugovk:
You're welcome! Python 3.5 failed, I guess it needs some backports the non-stdlib `mock` package, but will check back in a bit.

hugovk:
Updated and now ✅!

sloria:
@hugovk  Can you resolve the conflicts?

Once that's done and 6.0.0b1 is released, I think we're good to merge and release this cc @lafrech

hugovk:
Updated and now ✅!

src/webargs/core.py
```diff
@@ -1,11 +1,9 @@
...
-
 import functools
 import inspect
 import logging
 import warnings
 from copy import copy
+from collections.abc import Mapping
 
 try:
     import simplejson as json
```
lafrech: I suspect this could be removed.

lafrech (approved):
Looks good to me.

I suspect `simplejson` was used in PY2 to compensate for shortcomings in core `json`.

We should also test against 3.8. Let's do that in another PR.

sloria:
Yeah, the simplejson import can be removed, as it was only for Python 2 compat

https://github.com/marshmallow-code/webargs/blob/c90e814f4dc1624586cd04180f553a6c4c77f873/setup.py#L7-L8

hugovk:
Does that mean we should use universal wheels?

```diff
diff --git a/setup.cfg b/setup.cfg
 [bdist_wheel]
-# We build separate wheels for
-# Python 2 and 3 because of the conditional
-# dependency on simplejson
-universal = 0
+universal = 1
```

src/webargs/core.py
hugovk: Removed in b734c5508f63cea93f3f6cee34807f1d400d2fe9!

sloria:
Yes, good catch. We can build universal wheels now.

hugovk:
Updated!

lafrech:
Great. Let's move on with this.

lafrech:
Released as 6.0.0b2 !

Thanks again @hugovk.

hugovk:
You're welcome!

Issue #371, opened 2019-03-11
Race conditions for parallel requests due to cache


ThiefMaster:
I just noticed that something in webargs or marshmallow isn't thread-safe. Take this minimal example"

```python
import time

from flask import Flask, jsonify, request
from marshmallow.fields import Field
from webargs.flaskparser import use_kwargs


app = Flask(__name__)


class MyField(Field):
    def _serialize(self, value, attr, obj):
        return value

    def _deserialize(self, value, attr, data):
        print 'deserialize', request.json, value
        time.sleep(0.25)
        return value


@app.route('/test', methods=('POST',))
@use_kwargs({
    'value': MyField(),
})
def test(value):
    time.sleep(1)
    return jsonify(webargs_result=value, original_data=request.json['value'])
```

Run it with threading enabled:

    $ FLASK_APP=webargsrace.py flask run -p 8080 --with-threads

Now send two requests in parallel, with different values:

    $ http post http://127.0.0.1:8080/test 'value=foo' & ; http post http://127.0.0.1:8080/test 'value=bar' &

The output from these two requests is:

```json
{
    "original_data": "bar",
    "webargs_result": "bar"
}
```

```json
{
    "original_data": "foo",
    "webargs_result": "bar"
}
```

Clearly not what one would have expected! :bomb: 

The output of the `print` statement showing the request data and what the field receives confirms the issue:

```
deserialize {u'value': u'bar'} bar
deserialize {u'value': u'foo'} bar
```

---

Tested with the latest marshmallow/webargs from PyPI, and also the marshmallow3 rc (marshmallow==3.0.0rc4, webargs==5.1.2).

ThiefMaster:
I think this is caused by the fact that `self._cache` on the `Parser` is not thread-safe.

lafrech:
Thanks for the report and the test case.

Indeed your test passes when removing the caching mechanism.

sloria:
Thanks for reporting @ThiefMaster .

I won't have time to work on this for the next few days, but I'd gladly review/merge/release a patch. @ThiefMaster @lafrech Are either you up for this?

ThiefMaster:
I'm looking into fixing it right now

lafrech:
I'm investigating.

Removing the cache fixes the issue. I'm trying to see if there is a simple means to keep a cache with a scope restricted to the request.

[...]

PR #462, opened 2020-01-24
Remove the Parser cache

### LINKED ISSUE #374: Re-factor cache invalidation
https://github.com/marshmallow-code/webargs/issues/371#issuecomment-471578852

sirosen:
Because the cache is no longer used field-by-field to fetch data, there's significantly less value in keeping it. Combined with the fact that each parser instantiation was already clearing the cache to avoid a security bug ( #371 ), the cache is no longer actually used at all in most (any?) contexts.

Remove the cache and all of the machinery associated with it (Parser._clear_cache, Parser._clone, and relevant checks).

Resolves #374

---

If there's some case I'm not aware of in which the cache is really useful, that's fine. But based on my reading, I don't think it really does anything anymore, so I'm proposing its removal.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
+6.0.0b6 (Unreleased)
+********************
+
+Refactoring:
+
+* Remove the cache attached to webargs parsers. Due to changes between webargs
+  v5 and v6, the cache is no longer considered useful.
+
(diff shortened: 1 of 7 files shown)


lafrech (approved):

lafrech:
I think this is good to merge to. Do you mind doing the rebase?

sirosen:
Not at all. I'll have it done in just a second.

PR #463, opened 2020-01-25
Inject the location into ValidationErrors

### LINKED ISSUE #460: [RFC] Namespace error structures by location in response
I think there's a case we never really considered: two schemas in a different location with a common field name.

In webargs 5, this is not possible in a single `use_args` call. In a multi `use_args` call, it depends if the user specifies the locations. The response is ambiguous about which field is wrong.

In current implementation in webargs 6, the response is ambiguous. A way to fix this would be to namespace the error structures by location.

Just thinking out loud, writing this here before I forget.

sirosen:
In order to ensure that errors are always unambiguous, add the `location` to the message data for ValidationErrors.
This is done by adding a new level of nesting to the error messages, putting errors under the location as a new top-level key.

Resolves #460

---

I'm putting this in a PR to evaluate it, but I'm not 100% comitted to this solution. I'm going to ask a couple of questions on #460 to try to better understand what the options are.

### DIFF
--- a/tests/apps/flask_app.py
+++ b/tests/apps/flask_app.py
@@ -212,16 +212,16 @@ def echo_use_kwargs_missing(username, **kwargs):
 def handle_error(err):
     if err.code == 422:
         assert isinstance(err.data["schema"], ma.Schema)
-    try:
+
+    if MARSHMALLOW_VERSION_INFO[0] >= 3:
         return J(err.data["messages"]), err.code
+
     # on marshmallow2, validation errors for nested schemas can fail to encode:
     # https://github.com/marshmallow-code/marshmallow/issues/493
     # to workaround this, convert integer keys to strings
-    except TypeError:
+    def tweak_data(value):
+        if not isinstance(value, dict):
+            return value
+        return {str(k): v for k, v in value.items()}
 
-        def tweak_data(value):
-            if not isinstance(value, dict):
-                return value
-            return {str(k): v for k, v in value.items()}
-
-        return J({k: tweak_data(v) for k, v in err.data["messages"].items()}), err.code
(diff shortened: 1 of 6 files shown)


tests/apps/flask_app.py
```diff
@@ -212,4 +212,16 @@ def echo_use_kwargs_missing(username, **kwargs):
 def handle_error(err):
     if err.code == 422:
         assert isinstance(err.data["schema"], ma.Schema)
-    return J(err.data["messages"]), err.code
+    try:
+        return J(err.data["messages"]), err.code
+    # on marshmallow2, validation errors for nested schemas can fail to encode:
+    # https://github.com/marshmallow-code/marshmallow/issues/493
+    # to workaround this, convert integer keys to strings
+    except TypeError:
```
lafrech: I'd use `MARSHMALLOW_VERSION_INFO` here to

- ensure we are actually using MA2, not hiding another problem
- think of removing the code when dropping MA2

We can keep the try/except if you think it is better this way.

sirosen: I'll change it right now. I think being able to flag it easily when dropping MA2 support is reason enough.

Also, you're right that catching a `TypeError` could hide other things (in theory, at least).

sirosen:
Rebased so that this can merge cleanly. I think I said this clearly in #460, but to be clear: I no longer have any hesitation about merging this and putting it into 6.0.

lafrech:
Yes, I think this change makes total sense.

I guess it will mostly impact unit tests. I mean adapting to this change will probably be a few changes, if any, in application code, but mostly changes in unit tests asserting error messages. At least in my applications.

Let's go.

@sloria, I'd be happy to have your approval before releasing v6.

PR #464, opened 2020-01-25
Change DelimitedList to always go to/from a string

### LINKED ISSUE #423: RFC: Only accept delimited string in DelimitedList
`DelimitedList` accepts either a list or a delimited string (e.g. "foo,bar,baz").

I'd like to make it more strict by only accepting a delimited list. Rather than adding a `strict` parameter, I'm thinking of dropping the whole "also accept a list" feature.

Any reason to support both?

I understand it inherits from `List` because once the string is parsed, it can be deserialized as a normal list. But are there cases where you'd expect either a list or a delimited string?

sirosen:
Parsing/deserializing a DelimitedList will now insist that the input is a string or bytes. List-like values for DelimitedList are no longer valid.
Also, for symmetry, remove the `as_string` parameter -- always serialize DelimitedList as a string.

As a final change, make the default error message for an invalid delimited list slightly different from the one for List, to disambiguate things.

closes #423

---

I think this is a significant improvement. I don't like the way in v5 that you can have
`...?foo=1,2,3` parse to `foo = ["1", "2", "3"]`, and `...?foo=1&foo=2&foo=3` parse to `foo = ["1", "2", "3"]`, but (**!**) `...?foo=1,2&foo=3` parses as `foo = ["1,2", "3]`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+6.0.0b5 (Unreleased)
+********************
+
+Refactoring:
+
+* *Backwards-incompatible*: `DelimitedList` now requires that its input be a
+  string and always serializes as a string. It can still serialize and deserialize
+  using another field, e.g. `DelimitedList(Int())` is still valid and requires
+  that the values in the list parse as ints.
+
 6.0.0b4 (2020-01-28)
 ********************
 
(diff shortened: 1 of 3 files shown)


lafrech (approved):
Thanks.

sirosen:
I've rebased this to cleanup the changelog conflict. I'll do the same on my other PRs to keep them fresh, since I'm not sure what else to do with multiple things in flight at once.

PR #472, opened 2020-01-31
Use kw-only arguments where sensible

lafrech:
Closes the second item in #440.

### DIFF
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -66,7 +66,7 @@ def is_json(mimetype):
     return False
 
 
-def parse_json(s, encoding="utf-8"):
+def parse_json(s, *, encoding="utf-8"):
     if isinstance(s, bytes):
         try:
             s = s.decode(encoding)
@@ -125,7 +125,7 @@ class Parser:
         "json_or_form": "load_json_or_form",
     }
 
-    def __init__(self, location=None, error_handler=None, schema_class=None):
+    def __init__(self, location=None, *, error_handler=None, schema_class=None):
         self.location = location or self.DEFAULT_LOCATION
         self.error_callback = _callable_or_raise(error_handler)
         self.schema_class = schema_class or self.DEFAULT_SCHEMA_CLASS
@@ -152,7 +152,7 @@ class Parser:
             raise ValueError('Invalid location: "{}"'.format(location))
         return function
 
-    def _load_location_data(self, schema, req, location):
+    def _load_location_data(self, *, schema, req, location):
         """Return a dictionary-like object for the location on the given request.
 
         Needs to have the schema in hand in order to correctly handle loading
@@ -168,7 +168,7 @@ class Parser:
         return data
 
     def _on_validation_error(
-        self, error, req, schema, location, error_status_code, error_headers
+        self, error, req, schema, location, *, error_status_code, error_headers
     ):
         # rewrite messages to be namespaced under the location which created
         # them
@@ -177,7 +177,13 @@ class Parser:
         #      {"foo":["Not a valid integer."]}
         error.messages = {location: error.messages}
         error_handler = self.error_callback or self.handle_error
-        error_handler(error, req, schema, error_status_code, error_headers)
+        error_handler(
+            error,
+            req,
+            schema,
+            error_status_code=error_status_code,
+            error_headers=error_headers,
+        )
 
     def _validate_arguments(self, data, validators):
         for validator in validators:
@@ -201,7 +207,7 @@ class Parser:
         elif callable(argmap):
             schema = argmap(req)
         else:
-            schema = dict2schema(argmap, self.schema_class)()
+            schema = dict2schema(argmap, schema_class=self.schema_class)()
         if MARSHMALLOW_VERSION_INFO[0] < 3 and not schema.strict:
             warnings.warn(
                 "It is highly recommended that you set strict=True on your schema "
@@ -214,10 +220,11 @@ class Parser:
         self,
         argmap,
         req=None,
+        *,
         location=None,
         validate=None,
         error_status_code=None,
-        error_headers=None,
+        error_headers=None
     ):
         """Main request parsing method.
 
@@ -255,7 +262,12 @@ class Parser:
             self._validate_arguments(data, validators)
         except ma.exceptions.ValidationError as error:
             self._on_validation_error(
-                error, req, schema, location, error_status_code, error_headers
+                error,
+                req,
+                schema,
+                location,
+                error_status_code=error_status_code,
+                error_headers=error_headers,
             )
         return data
 
@@ -283,11 +295,12 @@ class Parser:
         self,
         argmap,
         req=None,
+        *,
         location=None,
         as_kwargs=False,
         validate=None,
         error_status_code=None,
-        error_headers=None,
+        error_headers=None
     ):
         """Decorator that injects parsed arguments into a view function or method.
 
@@ -316,7 +329,7 @@ class Parser:
         # Optimization: If argmap is passed as a dictionary, we only need
         # to generate a Schema once
         if isinstance(argmap, Mapping):
-            argmap = dict2schema(argmap, self.schema_class)()
+            argmap = dict2schema(argmap, schema_class=self.schema_class)()
 
         def decorator(func):
             req_ = request_obj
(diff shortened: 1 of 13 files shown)


lafrech:
Looks like this is not allowed in Python 3.5

```py
    def parse(
        self,
        argmap,
        req=None,
        *,
        location=None,
        validate=None,
        error_status_code=None,
        error_headers=None,
    ):
```

I can't find a documented change in 3.6 explaining this.

I could go with

```py
    def parse(
        self,
        argmap,
        *,
        req=None,
        location=None,
        validate=None,
        error_status_code=None,
        error_headers=None,
    ):
```

but wouldn't it be too much?

sirosen:
Just in case someone else thinks of it, I tried the "too clever solution" of making parse's signature different in different pythons, but it doesn't work easily because py3.5 can't even parse the desired signature for 3.6+. I'll share because it's fun to think about:
<details><summary>polyglot_parse decorator</summary>

```python
# decorate it to have different signatures
# doesn't work in py3.5
def polyglot_parse(real_parse_func):
    if sys.version_info < (3, 6):
        def parse(
            self,
            argmap,
            req=None,
            location=None,
            validate=None,
            error_status_code=None,
            error_headers=None,
        ):
            return real_parse_func(
                self,
                argmap,
                req=req,
                location=location,
                validate=validate,
                error_status_code=error_status_code,
                error_headers=error_headers,
            )
    else:
        def parse(
            self,
            argmap,
            req=None,
            *,
            location=None,
            validate=None,
            error_status_code=None,
            error_headers=None,
        ):
            return real_parse_func(
                self,
                argmap,
                req=req,
                location=location,
                validate=validate,
                error_status_code=error_status_code,
                error_headers=error_headers,
            )

    return parse
```
</details>

It is possible if we're determined -- you can do things like adding a second module which only gets imported in 3.6, or write it all as a string and use `exec`! But... let's not go there, right? It seems like a mess.

There's also the "faking it" way with a `*args` list:
```python
def parse(
    self,
    argmap,
    *args,
    req=None,
    location=None,
    validate=None,
    error_status_code=None,
    error_headers=None,
):
    if len(args) > 1:
        raise ValueError
    if args and req is not None:
        raise ValueError
    if args[0] is not None and req is None:
        req = args[0]
    ...
```

But this also seems awkward and undesirable.

Python 3.5 EOLs later this year, but I don't want to wait that long to be able to use webargs 6.

I'm not sure if there's a potential issue is with `req` being keyword-only, other than it not being the API we _want_ for `parse` (I agree with the proposed, ideal function signature, with `req` as an optional positional arg). Should we just do that, even though it's not perfect?

The major alternative I see is just leaving this as-is, on the grounds that we can't do what we really want. It can be changed in a later major version upgrade.

lafrech:
Thanks for looking into this. I don't want to overthink it. I just exposed it here in the hope to get an explanation because I couldn't find any item in the changelog explaining this.

I agree we don't want to do anything messy.

Either we decide `req` is keyword only. It means modifying a lot of calls in the tests. I don't mind doing this but I figured it might be less practical for client code calling `parse`. In practice the calls are probably centralized in decorators, not scattered among the whole client application code, so it's not that bad. Also, it is not coherent with the way `req` is treated in `handle_error`. But it is not that bad either.

Or we just skip this method.

No big deal either way. I'll settle this one way or another.

lafrech:
Ah ah. Turns out the issue was not exactly the kw-only thing but the trailing comma in the signature.

Apparently, the trailing comma is fine, unless there is a kw-only `*` in the signature.

So this works:

```py
    def parse(
        self,
        argmap,
        req=None,
        *,                  # <-----------
        location=None,
        validate=None,
        error_status_code=None,
        error_headers=None
    ):
        pass

    def parse(
        self,
        argmap,
        req=None,
        location=None,
        validate=None,
        error_status_code=None,
        error_headers=None,                  # <-----------
    ):
        pass
```
But this does not:

```py
    def parse(
        self,
        argmap,
        req=None,
        *,                  # <-----------
        location=None,
        validate=None,
        error_status_code=None,
        error_headers=None,                  # <-----------
    ):
        pass
```

Go figure...

(I just found an online Python 3.4 interpreter to test that.)

sirosen:
Yeah, I just checked and that works on my py35 version (3.5.6). And I upgraded to 3.5.9 to check, works on that too.
So I think it's safe to say that 3.5 in general supports this.

Weird though. _shrugs_

src/webargs/core.py
```diff
@@ -503,7 +515,7 @@ def load_files(self, req, schema):
         return missing
 
     def handle_error(
-        self, error, req, schema, error_status_code=None, error_headers=None
+        self, error, req, *, schema, error_status_code=None, error_headers=None
```
sirosen: Should `schema` be a required keyword-only arg?
I would have left all required arguments as positionals and made only the optional arguments (`error_status_code`, `error_headers`) keyword-only.
I also see some variation in the signatures for this in the various parsers, which I assume is accidental.

This is not critical, since `handle_error` is only really called internally and exposed as a hook for users to write something which gets called by webargs (if they write their own parser subclassing the core parser). But the two signatures which occur to me are
```python
# everything keyword-only
def handle_error(
    self, *, error, req, schema, error_status_code=None, error_headers=None
):
# optional args keyword-only
def handle_error(
    self, error, req, schema, *, error_status_code=None, error_headers=None
):
```

lafrech:
Thanks for reviewing. I pushed a new version.

I didn't find the signature variations but anyway, I checked everywhere and I think it is fine, now. I also changed the signature in docstrings/docs.

sirosen:
> I didn't find the signature variations but anyway, I checked everywhere and I think it is fine, now. I also changed the signature in docstrings/docs.

The thing which I see varying is whether or not `error_status_code, error_headers` are given default values of `None`.
Almost everywhere, I see these without defaults, but the core parser version and some of the testsuite versions use `error_status_code=None, error_headers=None`.

I don't think this really has any impact -- those values should always be supplied when an error handler is called -- but now that I've noticed it, I can't help but suggest making them all identical.

(If I'm reading correctly, this was actually slightly inconsistent before, with some locations declaring them as positional arguments and some as keyword arguments.)

lafrech:
Good catch

I reworked the PR to make these consistent.

`error_status_code` and `error_headers` now default to `None` in `parse` and `use_args` and are required in error handler (which as I agree does not make much difference since we know they are supplied).

sirosen (approved):
This looks good to me. I don't know if we need @sloria to weigh in on this or not.
(I am, after all, still just some guy on the Internet, not a maintainer. 😄 )

lafrech:
I'd like @sloria to approve the 6.0.0 release, be it before or after we merge this.

> I am, after all, still just some guy on the Internet, not a maintainer.

Good point. I just sent you an invitation to join the maintainers team.

sloria:
I'll take a look at the release when I have some time, but don't block on me! I consider @lafrech more the owner of webargs at this point. Also glad to see @sirosen as maintainer =).

As discussed on Slack, my only hesitation here is `req` as a kw-only argument. That should probably be positional, but I have low confidence in that opinion.

P.S. @sirosen email me at sloria1 AT gmail.com if you'd like to join the marshmallow slack.

tests/test_core.py
```diff
@@ -732,7 +738,7 @@ def factory(req):
             assert req is web_request
             return self.UserSchema(context={"request": req}, **strict_kwargs)
 
-        @parser.use_kwargs(factory, web_request)
+        @parser.use_kwargs(factory, req=web_request)
```
sloria: Maybe revert?

lafrech: Yep. I reverted all `use_args` calls but missed `use_kwargs`. Fixed.

lafrech:
I agree. I just edited the PR to make `req` positional. Thanks for your feedback on this.

sirosen:
I keep hesitating to merge this because the last couple of changes have been so small and subtle, but I think this is good now and I'm going to go ahead. Using my newfound powers!

I think this takes us down to "last call" for changes in 6.0.

lafrech:
> I think this takes us down to "last call" for changes in 6.0.

Yep. Let's release a last beta. Then it leaves us a few days to update the docs (upgrading instructions) during which clients can provide feedback about the beta.

lafrech:
I was about to write an upgrading guide, but on second thought, the comprehensive instructions in the CHANGELOG should be enough. I didn't see the point of duplicating them in another doc page.

PR #477, opened 2020-02-21
stdlib `wsgiref` requires argument for `read()`

suola:
PEP3333 says:

> A server should allow read() to be called without an argument, and
> return the remainder of the client's input stream.

However, Python standard library `wsgiref` does not allow calling `read`
without arguments. Falcon test client `falcon.testing.TestClient` uses
`wsgiref` - therefore content length should be passed an argument to
`read`.

### DIFF
--- a/tests/test_falconparser.py
+++ b/tests/test_falconparser.py
@@ -1,4 +1,5 @@
 import pytest
+import falcon.testing
 
 from webargs.testing import CommonTestCase
 from tests.apps.falcon_app import create_app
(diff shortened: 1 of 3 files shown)


suola:
Short example on how to reproduce:

```
import falcon
import falcon.testing
import webargs
from webargs.falconparser import use_kwargs
from marshmallow import Schema


class MySchema(Schema):
    name = webargs.fields.Str()


class Resource:
    @use_kwargs(MySchema)
    def on_post(self, req, resp):
        pass


api = falcon.API()
api.add_route("/", Resource())
client = falcon.testing.TestClient(api)
resp = client.simulate_post("/", json={"name": "richard feynman"})
```

lafrech:
I think you're right.

Could you please add a test for this?

And add yourself to AUTHORS file.

Thanks.

suola:
Added a test, please let me know if it's ok for webargs and/or if the patch needs anything else.

tests/test_falconparser.py
```diff
@@ -42,3 +43,10 @@ def test_invalid_json(self, testapp):
     def test_parsing_headers(self, testapp):
         res = testapp.get("/echo_headers", headers={"name": "Fred"})
         assert res.json == {"NAME": "Fred"}
+
+    # `falcon.testing.TestClient.simulate_request` parses request with `wsgiref`
+    def test_body_parsing_works_with_simulate(self, testapp):
```
lafrech: I don't think testapp is needed, here. Is it?

suola: You're right, it's not - updated.

lafrech:
Thanks.

PR #486, opened 2020-03-10
Fix location argument name for docs

LefterisJP:
It seems that the location argument name is wrong in the docs:
https://webargs.readthedocs.io/en/latest/api.html#webargs.core.Parser.use_kwargs

Probably a remnant of the pre v6.0.0 code. This should fix it.


lafrech:
Thanks. Please add yourself to AUTHORS file and we'll merge this.

LefterisJP:
@lafrech Thanks for the quick response. Done!

sirosen:
Yep, good catch!

Issue #483, opened 2020-02-29
Webargs 6.0.0 has broken Flaskparser @use_kwargs: Never parses query string or formdata


kirsle:
Hello,

Your recent release of 6.0.0 of webargs has unexpectedly broken our Flask apps. It seems now the @use_kwargs decorator from the `webargs.flaskparser` **completely fails to parse** GET query string parameters or POST form-data parameters. Only JSON request bodies ever get parsed anymore.

Here is an example Flask app that shows the problem:

```python
from flask import Flask, jsonify
from webargs import fields
from webargs.flaskparser import use_kwargs

app = Flask(__name__)

@app.route("/test1", methods=["GET", "POST", "PUT"])
@use_kwargs({
    "page": fields.Int(missing=1),
    "per_page": fields.Int(missing=20),
    "full": fields.Bool(missing=False),
})
def test1(**kwargs):
    return jsonify(kwargs)

@app.route("/test2", methods=["GET", "POST", "PUT"])
@use_kwargs({
    "name": fields.Str(required=True),
})
def test2(**kwargs):
    return jsonify(kwargs)

app.run()
```

The endpoint "/test1" has three optional parameters each with default values when missing. If I make a GET request (with query parameters) or a POST request with application/x-www-form-urlencoded format, webargs completely fails to pick up my parameters and only sees the defaults.

In the endpoint "/test2" I have a required parameter, which again fails to parse on GET or form-data post but works on JSON post only. Here are some curl examples:

```bash
# GET request doesn't parse any param
% curl 'http://localhost:5000/test1?page=5&per_page=2&full=1'
{"full":false,"page":1,"per_page":20}

# Normal POST doesn't parse any param
% curl -X POST -d page=5 -d per_page=2 -d full=1 'http://localhost:5000/test1'
{"full":false,"page":1,"per_page":20}

# JSON POST actually does parse params
% curl -X POST -H 'Content-Type: application/json' -d '{"page": 5, "per_page": 2, "full": true}' 'http://localhost:5000/test1'
{"full":true,"page":5,"per_page":2}

# GET is "missing required parameter"
% curl 'http://localhost:5000/test2?name=alice'
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>422 Unprocessable Entity</title>
<h1>Unprocessable Entity</h1>
<p>The request was well-formed but was unable to be followed due to semantic errors.</p>

# POST is "missing required parameter"
[...]

PR #489, opened 2020-03-17
Add an upgrading doc to help users switch to 6.0

### LINKED ISSUE #485: 6.0.0 requires changes to some pre_load hooks
Just upgraded to 6.0.0. Using `FlaskParser`, the input to `Schema.load` is now a `MultiDictProxy` wrapping a werkzeug `ImmutableMultiDict` instead of a regular dictionary. This breaks a schema I have that uses a `pre_load` hook to converts empty strings to `None`. (I'm processing an HTML form with a bunch of optional number inputs similar to some of the solutions in marshmallow-code/marshmallow#713)

Not the biggest deal to work around but I thought I would mention it because I didn't see it in the change log.

Also it would be convenient if `MultiDictProxy` had a `__repr__` to make debugging more convenient.

sirosen:
The doc covers several issues, and explicitly is meant to address and resolve #485.

The doc is modeled on the marshmallow upgrading doc, and documents every backwards incompatible change which was made in 6.0, along with an example of impacted code and a rewrite which restores equivalent or near-equivalent behavior.

The doc mostly steers clear of trying to explain the decisions made, but it does provide some context, especially at the beginning when we have to show major signature changes in `use_args`/`use_kwargs`/`parse`.
Most of the suggested rewrites are simple, but a couple -- most notably the UnpackingDelimitedListSchema case -- are ideas which might be non-obvious to users.

Except when documenting specific changes to `parse` and related functions, everything is using `locations=("query",)` to specify "query params only" in the 5.x version of the code. That way, the translation to 6.x remains very simple in these cases.

@lafrech, I remember you saying when we did the 6.0 release that you were going to skip on writing an upgrade guide, since the changelog is pretty comprehensive. I agreed, but after seeing #483 and #485 , I figured I'd take a couple of hours and knock this out.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -6,14 +6,10 @@ This section documents migration paths to new releases.
 Upgrading to 6.0
 ++++++++++++++++
 
-In webargs 6, the way that data is loaded from various locations and passed to
-your schemas changed significantly. This may require you to change several
-different parts of your webargs usage.
-
-Multiple locations are no longer supported in a single call
+Multiple Locations Are No Longer Supported In A Single Call
 -----------------------------------------------------------
 
-And only JSON data is parsed by default.
+The default location is JSON/body.
 
 Under webargs 5.x, you may have written code which did not specify a location.
 
@@ -29,23 +25,24 @@ parameter, you must be explicit and rewrite like so:
 
     # webargs 5.x
     @parser.use_args({"q": ma.fields.String()})
-    def foo(q):
-        return some_function(user_query=q)
+    def foo(args):
+        return some_function(user_query=args.get("q"))
 
 
     # webargs 6.x
     @parser.use_args({"q": ma.fields.String()}, location="query")
-    def foo(q):
-        return some_function(user_query=q)
+    def foo(args):
+        return some_function(user_query=args.get("q"))
 
 This also means that another usage from 5.x is not supported. If you had code
 with multiple locations in a single `use_args`, `use_kwargs`, or `parse` call,
-you must write it in multiple separate `use_args` invocations, like so:
+you must write it in multiple separate `use_args` or `use_kwargs` invocations,
+like so:
 
 .. code-block:: python
 
     # webargs 5.x
-    @parser.use_args(
+    @parser.use_kwargs(
         {
             "q1": ma.fields.Int(location="query"),
             "q2": ma.fields.Int(location="query"),
@@ -58,13 +55,13 @@ you must write it in multiple separate `use_args` invocations, like so:
 
 
     # webargs 6.x
-    @parser.use_args({"q1": ma.fields.Int(), "q2": ma.fields.Int()}, location="query")
-    @parser.use_args({"h1": ma.fields.Int()}, location="headers")
+    @parser.use_kwargs({"q1": ma.fields.Int(), "q2": ma.fields.Int()}, location="query")
+    @parser.use_kwargs({"h1": ma.fields.Int()}, location="headers")
     def foo(q1, q2, h1):
         ...
 
 
-Fields no longer support location=...
+Fields No Longer Support location=...
 -------------------------------------
 
 Because a single `parser.use_args`, `parser.use_kwargs`, or `parser.parse` call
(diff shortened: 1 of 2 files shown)


docs/upgrading.rst
```diff
@@ -0,0 +1,447 @@
...
+
+This section documents migration paths to new releases.
+
+Upgrading to 6.0
+++++++++++++++++
+
+In webargs 6, the way that data is loaded from various locations and passed to
+your schemas changed significantly. This may require you to change several
+different parts of your webargs usage.
+
```
lafrech: I tried to reword this, but in fact, I'd just remove it.

Next paragraphs speak for themselves.

docs/upgrading.rst
```diff
@@ -0,0 +1,447 @@
...
+Multiple locations are no longer supported in a single call
```
lafrech: I think the convention is to capitalize titles.

docs/upgrading.rst
```diff
@@ -0,0 +1,447 @@
...
+-----------------------------------------------------------
+
+And only JSON data is parsed by default.
```
lafrech: The default location is JSON/body.

docs/upgrading.rst
```diff
@@ -0,0 +1,447 @@
...
+    @parser.location_loader("data")
+    def load_data(request, schema):
+        return request.data
+
+Data is not filtered before being passed to your schema, and it may be proxified
+--------------------------------------------------------------------------------
+
+In webargs 5.x, the schema you gave was used to pull data out of the request
+object. That data was compiled into a dictionary which was then passed to your
+schema.
```
lafrech: In webargs 5.x, the deserialization schema was used to pull data out of the request
object. That data was compiled into a dictionary which was then passed to the schema.

docs/upgrading.rst
```diff
@@ -0,0 +1,447 @@
...
+
+One of the major changes in webargs 6.x allows the use of `unknown` parameter
+on schemas. This lets a schema decide what to do with fields not specified in
+the schema. In order to achieve this, webargs now passes the full data from
+the location you selected to your schema.
```
lafrech: webargs now passes the full data from specified location to the schema.

docs/upgrading.rst
```diff
@@ -0,0 +1,447 @@
...
+    # webargs 5.x
+    @use_args({"q": ma.fields.String()}, locations=("query",))
+    def viewfunc(q, request):
+        ...
+
+
+    # webargs 6.x
+    @use_args({"q": ma.fields.String()}, location="query")
+    def viewfunc(request, q):
+        ...
```
lafrech, line 476: I didn't double-check but I think request was always first and this change only affects multiple calls to `use_args`.

lafrech (commented):
Thank you so much for this.

I made a few comments. Otherwise, good to go!

Also, I generally prefer to avoid addressing the reader directly, using "you", using "one" instead. Merely a matter of style and I don't really mind. This doc is too precious to block on this.

docs/upgrading.rst
sirosen: 👍 I'll remove it in a fixup commit.

docs/upgrading.rst
sirosen: Okay, I'll switch all of the titles to Title Case.

docs/upgrading.rst
sirosen: I take it this is a suggested rewording? I'll go with it. It's shorter than what I had.

docs/upgrading.rst
sirosen: I'll check this and rewrite if I got it wrong.

sirosen:
I like all of your rewording. I'll do all of these except for the pyramid one right now -- I want to check the behavior and make sure I have a correct example of the change.

sirosen:
I've just gotten to testing the PyramidParser behaviors and pushing the fix. I gave a new example in which there's a user-defined decorator passing additional data, so it clearly shows how `use_args` changes ordering.

I also noticed that the `use_args` calls in this doc were using a kind of confusing convention of naming the arg dict after the only field used, like
```python
@use_args({"q": fields.String()})
def foo(q):
    bar(q)
```

To make the behavior of `use_args` less confusing, I changed all of these to
```python
@use_args({"q": fields.String()})
def foo(args):
    bar(args.get("q"))
```

lafrech:
Sorry, I forgot about this.

PR #495, opened 2020-04-10
[WIP] Improve validation error handling for flaskparser

reinhard-mueller:
This pull request makes the default validation error handling more useable out-of-the-box, and at the same time makes it easier to implement custom error handling. Features are:
* Full information about the validation error on the default error page returned to the client.
* If `error_headers` are passed to the `parse()` method, actually include them in the HTTP response even when using the default error handler.
* Common base class `VerboseError` for all HTTPException instances carrying verbose validation error information, so these errors can specifically be called with constructs like `except VerboseError:` or `@app.errorhandler(VerboseError)`.
* API upwards compatible with the previous behaviour.

Please let me know what you think about this. If you agree to include this patch, I will complete it by adding comments and test cases and updating documentation and examples.

### DIFF
--- a/src/webargs/flaskparser.py
+++ b/src/webargs/flaskparser.py
@@ -19,25 +19,66 @@ Example: ::
         return 'Hello ' + args['name']
 """
 import flask
-from werkzeug.exceptions import HTTPException
+from werkzeug.exceptions import default_exceptions
+from werkzeug.utils import escape
 
 from webargs import core
 from webargs.compat import MARSHMALLOW_VERSION_INFO
 from webargs.multidictproxy import MultiDictProxy
 
 
+class VerboseError:
+    """Mixin class for handling verbose error information and additional
+    headers.
+    """
+
+    def __init__(self, exc, **kwargs):
+        # Note that this will call the __init__() of the sibling class to which
+        # it is mixed in!
+        super().__init__()
+        self.exc = exc
+        self.messages = kwargs.get("messages")
+        self.extra_headers = kwargs.get("headers")
+        self.data = kwargs
+
+    def get_description(self, environ):
+        if self.messages is None:
+            return super().get_description()
+        html = []
+        for source, errors in self.messages.items():
+            html.append(f"<h2>Errors in {escape(source)}</h2>")
+            html.append("<p>")
+            if isinstance(errors, dict):
+                for field, message in errors.items():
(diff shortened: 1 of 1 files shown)


reinhard-mueller:
Just to avoid misunderstandings: this PR is marked "WIP" because I'm waiting for feedback.

sloria:
This seems like it belongs in a custom error handler. See https://webargs.readthedocs.io/en/latest/quickstart.html#error-handling . We don't want to be returning HTML, since the presentation of the errors is the responsibility of the client.

Thanks anyway for working on this.

reinhard-mueller:
Hi,

I'm not sure I understand your comment:
> We don't want to be returning HTML,

If I am not mistaken, the currently used exception class returns HTML in `get_dscription()` as well, with the only differentce that content of the HTML in the current exception class is much less informative.

sloria:
Ah, my apologies--I misread this. I wasn't aware of `get_description()` before. Is it only relevant in a debugging context?

Reopening, but unfortunately I don't have much time to review this right now. Going to defer to @lafrech or @sirosen on this.

reinhard-mueller:
It could be relevant in any contexts where you don't define a custom error handler - be it for debugging, be it in user-interactive environments where a HTML error response is desired and thus Flask's default error handling is sufficient.

sirosen:
> Just to avoid misunderstandings: this PR is marked "WIP" because I'm waiting for feedback.

Thanks for clarifying!
I hadn't really taken a look yet because I take "WIP" to mean it's too early for anyone to look unless they have spare time to provide early feedback.

---

I'm not sure I follow what's being said about `get_description`. In terms of separation between webargs and flask/werkzeug, that appears to me to be something new for webargs, added in this PR. Right now, the default handler is just calling `flask.abort` and letting that take care of things, without thinking or worrying about what that might mean.
I'm not aware of any place where webargs itself gets involved in the presentation of errors directly.

This really does seem to me like something which shouldn't be in webargs, but somewhere outside, as a user-responsibility. Not because I don't like nice error handlers, but because we want to keep the default error handler very lightweight and minimally invasive.

I'm not closing this yet, in case @lafrech disagrees, but I'm disinclined to accept the change at present.

lafrech:
I'm in line with the other comments. Besides, this is easy to achieve in user code.

Thanks for sharing, anyway.

PR #498, opened 2020-04-10
Add URL variable parameter to flaskparser example

### LINKED ISSUE #497: Suggest including URL variable parameters in more examples
I'm totally new to Flask, Marshmallow, SQLAlchemy, webargs and this ecosystem. I was given a coding test and while trying to complete it, I had to use flaskparser together with an URL parameter. Not knowing how to configure them to be used together, I googled "flaskparser". The top 5 results are:

- https://webargs.readthedocs.io/en/latest/_modules/webargs/flaskparser.html
- https://webargs.readthedocs.io/en/latest/
- https://webargs.readthedocs.io/en/latest/api.html
- https://webargs.readthedocs.io/en/latest/advanced.html
- https://webargs.readthedocs.io/en/latest/quickstart.html

None of these include an example where an URL variable parameter and webargs are used together. I found the example that I need in [Framework Support](https://github.com/marshmallow-code/webargs/blob/dev/docs/framework_support.rst) (which does not appear on the first page of the Google results) as following:

```
from webargs import fields
from webargs.flaskparser import use_args


@app.route("/user/<int:uid>")
@use_args({"per_page": fields.Int()}, location="query")
def user_detail(args, uid):
    return ("The user page for user {uid}, showing {per_page} posts.").format(
        uid=uid, per_page=args["per_page"]
    )
```

Since the simpler use can can be inferred from the more complex use case, but not vice-versa, I would suggest to either include this code example in [flaskparser.py example](https://github.com/marshmallow-code/webargs/edit/dev/src/webargs/flaskparser.py), or somehow tweak the SEO for https://webargs.readthedocs.io/en/latest/framework_support.html so that it will appear on the first page of Google search results.

ugultopu:
Fixes #497. Copied the example in [Framework Support documentation](https://github.com/marshmallow-code/webargs/blob/dev/docs/framework_support.rst) to [flaskparser.py](https://github.com/marshmallow-code/webargs/blob/dev/src/webargs/flaskparser.py).

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -49,3 +49,4 @@ Contributors (chronological)
 * Sami Salonen `@suola <https://github.com/suola>`_
 * Tim Gates `@timgates42 <https://github.com/timgates42>`_
 * Lefteris Karapetsas `@lefterisjp <https://github.com/lefterisjp>`_
+* Utku Gultopu `@ugultopu <https://github.com/ugultopu>`_
--- a/src/webargs/flaskparser.py
+++ b/src/webargs/flaskparser.py
@@ -9,14 +9,16 @@ Example: ::
 
     app = Flask(__name__)
 
-    hello_args = {
-        'name': fields.Str(required=True)
+    user_detail_args = {
+        'per_page': fields.Int()
     }
 
-    @app.route('/')
-    @use_args(hello_args)
-    def index(args):
-        return 'Hello ' + args['name']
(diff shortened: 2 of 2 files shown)


lafrech:
Fine with me.

I don't feel the need to adapt the documentation to Google (c) ranking and the example is at its right place in the framework support docs. However, it also makes sense to put it here.

Please add yourself to AUTHORS.rst.

PR #506, opened 2020-04-29
add unknown optional keywords to parse and use_args

zenglanmu:
Allow user to set the unknown field is important in some cases. For example, the query also have an additional per_page args which handled by other library. Current webargs with version 6.0 behavior ould raise an validation error, which causes broken changes since verision 5. Maybe it is better to set unknown=EXCLUDE as default behavior or at least gives the user the ability to set it.

### DIFF
--- a/src/webargs/asyncparser.py
+++ b/src/webargs/asyncparser.py
@@ -30,7 +30,8 @@ class AsyncParser(core.Parser):
         location: str = None,
         validate: Validate = None,
         error_status_code: typing.Union[int, None] = None,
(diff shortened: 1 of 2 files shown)


sirosen:
As mentioned in #514, I think we prefer an approach in which dict schemas and regular schemas are treated the same.
This is going to be solved in v7.0 in a way that also handles setting defaults per location (see #544 for detail).

In light of that, and given that I'm trying to get 7.0 ready in the next few weeks or months, I'm going to close this.

PR #508, opened 2020-04-29
Remove all use of 2nd person from ugprade doc

sirosen:
This was an issue with the first version of this doc, but we decided to push it out without fixing it yet.

Using the second person in docs is a bad habit of mine. I just took a few minutes to go through and do the necessary rewrites.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -11,15 +11,15 @@ Multiple Locations Are No Longer Supported In A Single Call
 
 The default location is JSON/body.
 
-Under webargs 5.x, you may have written code which did not specify a location.
+Under webargs 5.x, code often did not have to specify a location.
 
-Because webargs would parse data from multiple locations automatically, you did
-not need to specify where a parameter, call it `q`, was passed.
+Because webargs would parse data from multiple locations automatically, users
+did not need to specify where a parameter, call it `q`, was passed.
 `q` could be in a query parameter or in a JSON or form-post body.
 
-Now, webargs requires that you specify only one location for data loading per
+Now, webargs requires that users specify only one location for data loading per
 `use_args` call, and `"json"` is the default. If `q` is intended to be a query
-parameter, you must be explicit and rewrite like so:
+parameter, the developer must be explicit and rewrite like so:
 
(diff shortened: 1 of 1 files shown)


docs/upgrading.rst
```diff
@@ -34,9 +34,9 @@ parameter, you must be explicit and rewrite like so:
     def foo(args):
         return some_function(user_query=args.get("q"))
 
-This also means that another usage from 5.x is not supported. If you had code
-with multiple locations in a single `use_args`, `use_kwargs`, or `parse` call,
-you must write it in multiple separate `use_args` or `use_kwargs` invocations,
+This also means that another usage from 5.x is not supported. Dode with
```
lafrech: Code

docs/upgrading.rst
```diff
@@ -118,15 +120,14 @@ the schema. In order to achieve this, webargs now passes the full data from
...
-types which can return one or multiple values as you prefer. To handle
-`marshmallow.fields.List` and `webargs.fields.DelimitedList` fields correctly,
-passing list data, webargs must combine schema information with the raw request
-data. This is done in the
+types which can return one or multiple values. To handle `marshmallow.fields.List`
+and `webargs.fields.DelimitedList` fields correctly, passing list data, webargs
+must combine schema information with the raw request data. This is done in the
 :class:`MultiDictProxy <webargs.multidictproxy.MultiDictProxy>` type, which
-will often be passed to your schemas.
+will often be passed to schemas.
```
lafrech: Should the whole paragraph be moved to line 180, with "Finally, this change passes a proxy object where schemas once saw a dict."?

Looks like this chapter interleaves unknown and multidict issues. Or maybe I read too fast.

lafrech (approved):
Just two comments.

Looks good otherwise.

👍

sirosen:
I've made a couple of changes for review and pushed in a new commit to make it easy to see.

I agree with moving that paragraph about MultiDictProxy down to go with the doc on how to adapt usage. Looking at that section, it's all about "making changes to data passed to schemas", so it holds together logically. We could split it up if you want, but I think it's okay?

PR #509, opened 2020-04-29
Add DelimitedTuple field (ma3 only)

### LINKED ISSUE #502: 'Not a valid tuple.'  when trying to use marshmallow fields.Tuple for argument validation
I'm trying to use the marshmallow fields.Tuple for querystring argument validation on a GET request using Flask.  The issue I'm running into is that no matter what type of object I declare and no matter what I use in the request, I always get the default 'Not a valid tuple.' response.  I have tried using a tuple of size 1 and 2; using fields.String and/or fields.Integer, etc with the same result.

- I'm using Python 3.6.9 with these dependencies:
anyjson==0.3.3
apipkg==1.5
arrow==0.15.5
attrs==19.3.0
backports.functools-lru-cache==1.6.1
cassandra-driver==3.22.0
Cerberus==1.3.2
certifi==2019.11.28
cffi==1.13.2
chardet==3.0.4
click==7.1.1
execnet==1.7.1
Flask==1.1.1
Flask-Cors==3.0.8
funcsigs==1.0.2
futures==3.1.1
geomet==0.1.2
gevent==1.4.0
greenlet==0.4.13
gunicorn==20.0.4
idna==2.9
importlib-metadata==1.6.0
itsdangerous==1.1.0
Jinja2==2.11.1
jsonklog==0.15.0
MarkupSafe==1.1.1
marshmallow==3.5.1
neurolab==0.3.5
numpy==1.18.1
pluggy==0.13.1
py==1.8.1
pyaml==20.3.1
pymongo==3.10.1
pytest==3.3.0
pytest-forked==0.2
pytest-xdist==1.20.1
python-dateutil==2.8.1
PyYAML==5.3.1
readline==6.2.4.1
requests==2.23.0
six==1.14.0
urllib3==1.25.8
webargs==6.0.0
Werkzeug==1.0.0
zipp==3.1.0

- Here is an example of what I'm trying to do:
```
from flask import Flask
from webargs.flaskparser import parser, use_kwargs
from marshmallow import EXCLUDE, fields, Schema


app = Flask(__name__)


@app.errorhandler(422)
def custom_handler(error):
    errors = []
    if 'query' in error.data['messages']:
        for arg in error.data['messages']['query']:
            for item in error.data['messages']['query'][arg]:
                errors.append(item)
    return str(errors), 400


class test_schema(Schema):
    class Meta:
        unknown = EXCLUDE
        strict = True
    
    test_tup = fields.Tuple((fields.Integer(required=True), fields.Integer(required=True)), required=True)


@app.route('/test/', strict_slashes=False)
@parser.use_kwargs(test_schema, location='query')
def test_the_mallow(**kwargs):
    return "True"
```

- Finally, here are a couple example url's I've tried:
  localhost:2300/test/?test_tup=[0,0]
  localhost:2300/test/?test_tup=(0,0)
  localhost:2300/test/?test_tup=0,0

sirosen:
When running with marshmallow3, ma.fields.Tuple provides a basis for a second type, similar to DelimitedList. Generalize most of DelimitedList into a mixin class which describes the desired way of de/serializing with a delimiter, in terms of some (unknown) parent class. DelimitedList and DelimitedTuple are just instances of this.

Apply the DelimitedList tests to DelimitedTuple.

closes #502

---

I'm happy to share code between these classes some other way, not using the mixin. I thought for a second about building a factory function that could be used like
```python
DelimitedList = make_delimited_variant(ma.fields.List)
```
but I think this way is simpler.

### DIFF
--- a/src/webargs/fields.py
+++ b/src/webargs/fields.py
@@ -43,7 +43,40 @@ class Nested(ma.fields.Nested):
         super().__init__(nested, *args, **kwargs)
 
 
-class DelimitedList(ma.fields.List):
+class DelimitedFieldMixin:
+    """
+    This is a mixin class for subclasses of ma.fields.List and ma.fields.Tuple
+    which split on a pre-specified delimiter. By default, the delimiter will be ","
+
+    Because we want the MRO to reach this class before the List or Tuple class,
+    it must be listed first in the superclasses
+
+    For example, a DelimitedList-like type can be defined like so:
+
+    >>> class MyDelimitedList(DelimitedFieldMixin, ma.fields.List):
+    >>>     pass
+    """
+
+    delimiter = ","
+
+    def _serialize(self, value, attr, obj):
+        # serializing will start with parent-class serialization, so that we correctly
+        # output lists of non-primitive types, e.g. DelimitedList(DateTime)
+        return self.delimiter.join(
+            format(each) for each in super()._serialize(value, attr, obj)
+        )
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        # attempting to deserialize from a non-string source is an error
+        if not isinstance(value, (str, bytes)):
+            if MARSHMALLOW_VERSION_INFO[0] < 3:
+                self.fail("invalid")
+            else:
+                raise self.make_error("invalid")
+        return super()._deserialize(value.split(self.delimiter), attr, data, **kwargs)
+
+
+class DelimitedList(DelimitedFieldMixin, ma.fields.List):
     """A field which is similar to a List, but takes its input as a delimited
     string (e.g. "foo,bar,baz").
 
@@ -61,18 +94,25 @@ class DelimitedList(ma.fields.List):
         self.delimiter = delimiter or self.delimiter
         super().__init__(cls_or_instance, **kwargs)
 
-    def _serialize(self, value, attr, obj):
-        # serializing will start with List serialization, so that we correctly
-        # output lists of non-primitive types, e.g. DelimitedList(DateTime)
-        return self.delimiter.join(
-            format(each) for each in super()._serialize(value, attr, obj)
-        )
 
-    def _deserialize(self, value, attr, data, **kwargs):
-        # attempting to deserialize from a non-string source is an error
-        if not isinstance(value, (str, bytes)):
-            if MARSHMALLOW_VERSION_INFO[0] < 3:
-                self.fail("invalid")
-            else:
-                raise self.make_error("invalid")
-        return super()._deserialize(value.split(self.delimiter), attr, data, **kwargs)
+# DelimitedTuple can only be defined when using marshmallow3, when Tuple was
+# added
+if MARSHMALLOW_VERSION_INFO[0] >= 3:
+
+    class DelimitedTuple(DelimitedFieldMixin, ma.fields.Tuple):
+        """A field which is similar to a Tuple, but takes its input as a delimited
+        string (e.g. "foo,bar,baz").
+
+        Like Tuple, it can be given a tuple of nested field types which it will use to
+        de/serialize each element of the tuple.
+
+        :param Iterable[Field] tuple_fields: An iterable of field classes or instances.
+        :param str delimiter: Delimiter between values.
+        """
+
+        default_error_messages = {"invalid": "Not a valid delimited tuple."}
+        delimiter = ","
+
+        def __init__(self, tuple_fields, *, delimiter=None, **kwargs):
+            self.delimiter = delimiter or self.delimiter
+            super().__init__(tuple_fields, **kwargs)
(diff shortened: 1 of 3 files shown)


src/webargs/fields.py
```diff
@@ -76,3 +74,45 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+if MARSHMALLOW_VERSION_INFO[0] >= 3:
+
+    class DelimitedTuple(DelimitedFieldMixin, ma.fields.Tuple):
+        """A field which is similar to a Tuple, but takes its input as a delimited
+        string (e.g. "foo,bar,baz").
+
+        Like Tuple, it can be given a tuple of nested field types which it will use to
+        de/serialize each element of the tuple.
+
+        :param Field cls_or_instance: A field class or instance.
```
lafrech: -> tuple_fields

tests/test_core.py
```diff
@@ -869,6 +900,31 @@ def test_delimited_list_as_string_v2(web_request, parser):
...
+    schema = schema_cls()
+
+    parsed = parser.parse(schema, web_request)
+    assert parsed["dates"] == (
+        datetime.datetime(2018, 11, 1),
+        datetime.datetime(2018, 11, 2),
+    )
+
+    dumped = schema.dump(parsed)
+    data = dumped.data if MARSHMALLOW_VERSION_INFO[0] < 3 else dumped
```
lafrech: Useless since test only runs under MA3

tests/test_core.py
```diff
@@ -851,6 +851,37 @@ def test_delimited_list_default_delimiter(web_request, parser):
...
+    schema_cls = dict2schema(
+        {"ids": fields.DelimitedTuple((fields.Int, fields.Int, fields.Int))}
+    )
+    schema = schema_cls()
+
+    with pytest.raises(ValidationError):
+        parser.parse(schema, web_request)
+
+
 def test_delimited_list_as_string_v2(web_request, parser):
```
lafrech: Not sure why the v2 in the name. (Unrelated to this PR but could be fixed in it anyway.)

tests/test_core.py
```diff
@@ -869,6 +900,31 @@ def test_delimited_list_as_string_v2(web_request, parser):
     assert data["dates"] == "2018-11-01,2018-11-02"
 
 
+@pytest.mark.skipif(
+    MARSHMALLOW_VERSION_INFO[0] < 3, reason="fields.Tuple added in marshmallow3"
+)
+def test_delimited_tuple_as_string(web_request, parser):
```
lafrech: I don't understand the purpose of the "as_string" tests.

What's in there than is not tested in the default delimiter test above?

lafrech (commented):
Thanks.

Looks good.

Mixin approach is the best IMHO.

tests/test_core.py
sirosen: Yeah, I'll slip that into my cleanup. I couldn't tell you why either -- maybe there used to be two such tests?

tests/test_core.py
sirosen: I think this was something of a side-effect of #464 (also explains the v2 in the other test name).
I'm looking right now to see if there's really a difference. I don't think there is.

sirosen: I compared things and this does sort of check that deserialization with `dump()` produces good results in a nontrivial case. I updated the docstring, rather than removing the tests, but I'd be comfortable removing them too. Let me know which way you prefer.

lafrech: I'd have been tempted to just merge the basic and the datetime test into one. TupleField allows to use both an IntField and a DateTimeField at once.

Having two tests is just as fine.

lafrech (approved):
Great. Thanks.

I suppose you added a new commit to help me reviewing but you'd like to squash before merging.

I can do the squash / update changelog / release later when I get the time.

sirosen:
> I suppose you added a new commit to help me reviewing but you'd like to squash before merging.

Yep! I'm happy to squash things together now. I just figured the added diff would make it easier.

> I can do the squash / update changelog / release later when I get the time.

I'm also happy to do this. As a relatively new contributor, I don't want to be too pushy about my changes or doing releases -- let me know if I'm being too tentative.

I'll do a rebase and add a changelog entry, so this should simply be a merge + tweak to the changelog (to add the release date) when the time comes.

lafrech:
No pb @sirosen.

As long as a change is approved, we can merge it. And we generally comment in the PR to say "I'll release" or "Let's finish this other PR before releasing".

The release process (update changelog, version, etc.)  is a bit of a chore, so no one will blame you for doing it.

It is indeed a good habit to include the changelog update in the PR to lower the burden on the merger. I prefer to do it in a new commit at the last moment, when I'm confident the PR is the next merge, to avoid merge issues.

This is complete so I'll merge it.

Feel free to release when you get the 5 or 10 minutes it takes.

sirosen:
I forgot about this all day. I'm going to go ahead and do a release of it as v6.1 with this included before I can forget again.

PR #512, opened 2020-05-09
Update docs to reflect current handle_error argument names

jaswilli:
Hi,

I recently did the 5 to 6 upgrade and after following the guide I was still having trouble:

`TypeError: handle_webargs_error() got an unexpected keyword argument 'error_status_code'`

Which after a little digging I realized was because the argument names to the error handler hook (`@parser.error_handler`) had changed and become keyword-only. It seems like the change was mostly overlooked in the places it appears in the docs.

This pull request just updates the docs to update the argument names and call out what's necessary in the 5 -> 6 transition.

Hopefully I didn't overlook anything obvious but I'm happy to be educated if I did.


sirosen:
Thanks for the catch! And I like the addition of a section to the upgrading guide, since the change to `Parser.handle_error` isn't enough on its own in there.

I just scanned over things again to see if we missed anything else but don't see anything.

Issue #507, opened 2020-04-29
[RFC] Default to unknown=EXCLUDE?


lafrech:
Question raised in #480.

`unknown=RAISE` may be a nice default in marshmallow but it can be cumbersome to change in webargs.

In practice it is mostly useful in body payload.

I wouldn't use it in query args, for instance, because by construction of the framewok, there may be several layers of decorators adding different fields to parse.

In flask-smorest, I added a [doc section](https://flask-smorest.readthedocs.io/en/latest/arguments.html#multiple-arguments-schemas) to explain it but it is still surprising for the user (see https://github.com/marshmallow-code/flask-smorest/issues/145).

While the base schema is the way to go in marshmallow and is convenient once you are familiar with this logic, it is not practical to with arguments defined with `dict2schema`.

I guess in practice, `dict2schema` is used mostly for arguments with a few fields (typically query args) while the schema form is adapted to huge schemas (typically json body). Using `EXCLUDE` as default would make the `dict2field` form work while the user could easily opt-in to `RAISE` when using the schema form.

Should we change the default to `EXCLUDE` in a major release and let the user opt-in to `RAISE`?

lafrech:
I'm realizing that my question above is based on the wrong assumption that the `unknown` parameter of the schema is set in webargs, whish is incorrect. When using a schema, webargs does not affect this parameter.

The question in #480 was about changing the default for `dict2schema` only.

I don't see a nice way to change the default in marshmallow schemas. Changing it inside user-provided schema sounds bad. Providing a base schema with `EXCLUDE` as default would not be of much help: it wouldn't be less cumbersome than defining one's own schema.

But having the default in dict2schema differ from Schema sounds bad, doesn't it?

zenglanmu:
at least there should be an option to set `EXCLUDE` when using dict_arg not user defined schema class. User defined schema classes  is fine because `unknown=EXCLUDE`  can be easily passed through `class Meta`. We can do some thing like:
`dict2schema(dct, schema_class)(unknown=unknown)`
where parameter `unknown` can be passed by user

lafrech:
That's the point of https://github.com/marshmallow-code/webargs/issues/480#issuecomment-620771720.

I wouldn't object to that.

Well, the thing is `dict2schema` is imported from marshmallow 3, which adds inertia. I'm not sure we want to change it there, or if we do, it might be part of a larger change because if we add `unknown`, why not add other parameters as well?

zenglanmu:
[...]

PR #514, opened 2020-05-13
Add ability to pass `unknown` in parse calls

sirosen:
This is half of #507 . For the other half, I think we need to wait for 7.0, so we can make a breaking change to the default behavior.

This adds support for passing the `unknown` parameter in two major locations: Parser instantiation, and Parser.parse calls. use_args and use_kwargs are just parse wrappers, and they need to pass it through as well.

It also adds support for a class-level default for unknown, `Parser.DEFAULT_UNKNOWN`, which sets `unknown` for any future parser instances.

Explicit tweaks to handle this were necessary in asyncparser and PyramidParser, due to odd method signatures.

Support is tested in the core tests, but not the various framework tests. (I wanted to add this, but couldn't think of an easy/good way to do it. Suggestions welcome.)

Add a 6.2.0 (Unreleased) changelog entry with detail on this change. The changelog states that we will change the DEFAULT_UNKNOWN default in a future major release. Presumably we'll make it `EXCLUDE`, but I'd like to make it location-dependent if feasible, so I didn't commit to anything in the phrasing.

---

This obsoletes #506 if merged. I like to incorporate people's work if possible, but I don't see a way in this case. I don't think we should make this behavior special for `dict2schema`. I'd rather add a general feature which happens to play nice with `dict2schema`. That way, you get fewer chances for surprising behavioral changes if you replace a dict with a full schema object.

My first attempt at this was much more complex and didn't come together well.
I wanted to have a mapping like
```
{
  "query": EXCLUDE,
  "json": RAISE,
  ...
}
```
but putting this together and making it user-settable was just a lot of work. I'd rather do this, which is simpler, and get it into people's hands if they're having issues moving to 6.x.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,51 @@
 Changelog
 ---------
 
+6.2.0 (Unreleased)
+******************
+
+Features:
+
+* Add a new ``unknown`` parameter to ``Parser.parse``, ``Parser.use_args``, and
+  ``Parser.use_kwargs``. When set, it will be passed to the ``Schema.load``
+  call. If set to ``None`` (the default), no value is passed, so the schema's
+  ``unknown`` behavior is used.
+
+This allows usages like
+
+.. code-block:: python
+
+    import marshmallow as ma
+
+    # marshmallow 3 only, for use of ``unknown`` and ``EXCLUDE``
+    @parser.use_kwargs(
+        {"q1": ma.fields.Int(), "q2": ma.fields.Int()}, location="query", unknown=ma.EXCLUDE
+    )
+    def foo(q1, q2):
+        ...
+
+* Add the ability to set defaults for ``unknown`` on either a Parser instance
+  or Parser class. Set ``Parser.DEFAULT_UNKNOWN`` on a parser class to apply a value
+  to any new parser instances created from that class, or set ``unknown`` during
+  ``Parser`` initialization.
+
+Usages are varied, but include
+
+.. code-block:: python
+
+    import marshmallow as ma
+    from webargs.flaskparser import FlaskParser
+
+    parser = FlaskParser(unknown=ma.INCLUDE)
+
+    # as well as...
+    class MyParser(FlaskParser):
+        DEFAULT_UNKNOWN = ma.INCLUDE
+
+
+    parser = MyParser()
+
+
 6.1.0 (2020-04-05)
 ******************
 
--- a/src/webargs/asyncparser.py
+++ b/src/webargs/asyncparser.py
@@ -9,6 +9,7 @@ from marshmallow import Schema, ValidationError
 from marshmallow.fields import Field
 import marshmallow as ma
 
+from webargs.compat import MARSHMALLOW_VERSION_INFO
 from webargs import core
 
 Request = typing.TypeVar("Request")
@@ -28,6 +29,7 @@ class AsyncParser(core.Parser):
         req: Request = None,
         *,
         location: str = None,
+        unknown: str = None,
         validate: Validate = None,
         error_status_code: typing.Union[int, None] = None,
         error_headers: typing.Union[typing.Mapping[str, str], None] = None
@@ -38,6 +40,10 @@ class AsyncParser(core.Parser):
         """
         req = req if req is not None else self.get_default_request()
         location = location or self.location
+        unknown = unknown or self.unknown
+        load_kwargs = (
+            {"unknown": unknown} if MARSHMALLOW_VERSION_INFO[0] >= 3 and unknown else {}
+        )
         if req is None:
             raise ValueError("Must pass req object")
         data = None
@@ -47,7 +53,7 @@ class AsyncParser(core.Parser):
             location_data = await self._load_location_data(
                 schema=schema, req=req, location=location
             )
-            result = schema.load(location_data)
+            result = schema.load(location_data, **load_kwargs)
             data = result.data if core.MARSHMALLOW_VERSION_INFO[0] < 3 else result
             self._validate_arguments(data, validators)
         except ma.exceptions.ValidationError as error:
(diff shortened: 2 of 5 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,54 @@
...
+    # as well as...
+    class MyParser(FlaskParser):
+        DEFAULT_UNKNOWN = ma.INCLUDE
+
+
+    parser = MyParser()
+
+
+NOTE: No default value is set for ``DEFAULT_UNKNOWN`` for now, as doing so would be
+a breaking change. A future major release will change this behavior.
```
lafrech: I wouldn't mention this. Let's see what we'll do. (And a default value is set: `None`)

lafrech (approved):
Implementation is perfect.

Thanks for this.

Yes, testing this in all frameworks looks cumbersome and I don't think it adds much more coverage once the principle is tested in core tests.

As I wrote in #507, ideally, I'd like to set EXCLUDE as default and override to RAISE for body schemas, but we can't do that (unless perhaps we inspect the Meta attribute of the schema and I'd rather not do that). My use case should be achievable if we add a `location to unknown mapping`. This mapping could be a class attribute:

```py
    #: Default value to use for 'unknown' on schema load
    DEFAULT_UNKNOWN = None
    #: Default values to use for 'unknown' on schema load depending on location (overrides DEFAULT_UNKNOWN)
    DEFAULT_UNKNOWN_BY_LOCATION = {}
```

Then in user code:

```py
    DEFAULT_UNKNOWN = ma.EXCLUDE
    DEFAULT_UNKNOWN_BY_LOCATION = {'json': ma.RAISE}
```

User wouldn't need to define all locations. Only those that differ from default.

```py
        self.unknown = unknown or self.DEFAULT_UNKNOWN_BY_LOCATION.get(self.location, self.DEFAULT_UNKNOWN)
```

This can even be done in a minor version.

----------------------------------------------

Next major may override defaults to

```py
    DEFAULT_UNKNOWN = ma.EXCLUDE
    DEFAULT_UNKNOWN_BY_LOCATION = {}
```

CHANGELOG.rst
sirosen: Yeah, I'll just remove the note. We can decide how we want to change things when and if we do.

lafrech:
We should add a note in the docs. It can be after the "location" section of the quickstart: https://webargs.readthedocs.io/en/latest/quickstart.html#request-locations.

lafrech:
Things might change in marshmallow regarding propagation but currently, unknown doesn't propagate. There's a PR to propagate unknown when passed as load kwarg, but not as Schema param (https://github.com/marshmallow-code/marshmallow/pull/1575). I'm not sure this PR will make it because it is not consistent (load != schema init).

Let's assume unknown does not propagate. Passing it only makes sense for schemas without nesting. This is generally the case for headers and query args. But not for body args.

My concern is that we might be relying on something that will work in most cases but fail sometimes:

- If people use nesting in headers, query (achievable with custom logic),...
- If people declare body schemas and rely on this new feature to make them exclude.

Shouldn't we settle on something in marshmallow before making a decision here?

I do realize this PR fixes the from_dict case and I like the fact the it works for any schema (from_dict or classic Schema). Except it does not really work for Schema since nesting is broken.

sirosen:
First up: I'll add docs to this PR later, hopefully today.

> Shouldn't we settle on something in marshmallow before making a decision here?

I'm not convinced that we should wait. Maybe we want marshmallow settled before working on something more significant, like `LOCATION_DEFAULT_UNKNOWN`. But right now, I'm thinking about users who want to write

```python
@use_args(
    {"myheader": fields.Str(data_key="X-My-Header")},
    location="headers",
    unknown=ma.EXCLUDE
)
```

webargs isn't used in this case for marshmallow features, but merely to extract data from a location like headers or query params.

I've said before that users can just write a schema class, but seeing people's usage more is making me rethink that. Supporting `unknown` with `dict2schema` is valuable because it removes some of the temptation to write views like
```python
@use_args(SomeDataSchema, location="json")
def foo(request, args):
    myheader = request.headers.get("X-My-Header")
    ...
```
where basic parsing is mixed both inside and outside of the view code.

---

On the other hand, we have users who will want to write
```python
@use_args(SomeSchema, unknown=ma.EXCLUDE)
```
and will be surprised by the results (not propagating to nested schemas correctly).

However, that's not really "webargs' problem" to solve. The same thing happens with
```python
@use_args(SomeSchema(unknown=ma.EXCLUDE))
```
already. I don't think by exposing an additional way to pass `unknown` around that we're making the situation worse.

---

I can see another solution to all of this. But I might have to gear up and do (my first!) marshmallow PR to make it happen.

If marshmallow is not going to propagate `unknown` by default, but adds a new parameter next to `unknown` to control it, e.g. `propagate_unknown=True/False`, then we should wait. Because I think we would want `parse(..., unknown=INCLUDE)` to translate down into `schema.load(..., unknown=INCLUDE, propagate_unknown=True)`.

I'll post on ~that PR~ https://github.com/marshmallow-code/marshmallow/issues/1490 to suggest this. Maybe I can manage the PR this weekend.

sirosen:
I think I'd like to revisit merging this in its current state and doing a release.

> Shouldn't we settle on something in marshmallow before making a decision here?

I think https://github.com/marshmallow-code/marshmallow/pull/1634 is probably the best approach because it's backwards compatible and explicit. Maybe that's a bit egotistical, but hey, I wouldn't have submitted the PR if I didn't think it was a good idea! 😄 

But even if that PR were taken, I'm not sure how it impacts the behavior here. Would we really want to pass `propagate_unknown=True` by default? It feels a little bit like we're doing a bit too much if we do so.

But the biggest argument against it: it would mean that webargs behavior varies between different versions of marshmallow 3.
I don't like the idea of having behavior change subtly between `marshmallow 3.7` and `3.8` if we can avoid it.

Should I consider going ahead and merging this as-is? And when/if marshmallow 3 changes behavior, we can support that too, in the same way?

sirosen:
This is already included in the 7.0 work, so I'm closing this as a duplicate.

I don't think marshmallow is going to change this behavior, and I think we've accepted that. But taking care of this feature in a major version bump lets us solve the problem better and more thoroughly.

PR #525, opened 2020-06-10
Add a warning if validation errors are suppressed

### LINKED ISSUE #518: Parsing arguments always returns None
parsing the arguments always returns None, even if trying the most basic syntax taken from the [quick start page](https://webargs.readthedocs.io/en/latest/quickstart.html#basic-usage). Maybe I have a typo I failed to see.

Versions:
Flask: 1.1.2
Marshmallow: 3.6.0
Flask-marshmallow: 0.12
webargs: 6.1.0

Code:
```python
from pprint import pprint

from flask import request
from flask_restful import Resource
from webargs import fields, validate
from webargs.flaskparser import parser

# Flask-restful resource, GET method called with ?type=99999
class WebsiteList(Resource):
    @classmethod
    # @use_kwargs(website_args)
    def get(cls):
        website_args = {
            'type': fields.String(
                required=True,
                validate=validate.Range(min=0, max=6, error="testing error, never triggered")
            ),
        }

        # Line below outputs: ImmutableMultiDict([('type', '6')])
        pprint(request.args)

        args = parser.parse(website_args, request)
        # Line below outputs: None
        pprint(args)
```

I have also tried the decorator @use_kwargs(website_args) (And define website_args beforehand of course), but I get a traceback for TypeError: 'NoneType' object is not iterable:
```python
Traceback (most recent call last):
  File "C:\Users\Marco\.virtualenvs\api-kHflOUBu\lib\site-packages\flask\app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "C:\Users\Marco\.virtualenvs\api-kHflOUBu\lib\site-packages\flask\app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "C:\Users\Marco\.virtualenvs\api-kHflOUBu\lib\site-packages\flask_restful\__init__.py", line 468, in wrapper
    resp = resource(*args, **kwargs)
  File "C:\Users\Marco\.virtualenvs\api-kHflOUBu\lib\site-packages\flask\views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "C:\Users\Marco\.virtualenvs\api-kHflOUBu\lib\site-packages\flask_restful\__init__.py", line 583, in dispatch_request
    resp = meth(*args, **kwargs)
  File "C:\Users\Marco\.virtualenvs\api-kHflOUBu\lib\site-packages\webargs\core.py", line 364, in wrapper
    args, kwargs, parsed_args, as_kwargs
  File "C:\Users\Marco\.virtualenvs\api-kHflOUBu\lib\site-packages\webargs\core.py", line 298, in _update_args_kwargs
    kwargs.update(parsed_args)
TypeError: 'NoneType' object is not iterable
```

sirosen:
I considered having this raise an exception instead, but we can't be certain that the user doesn't know what they're doing in this case. It's conceivable that they have a validation hook which doesn't raise an error and they're prepared for `parse()` to return `None`. However, in that case, it would be beneficial to at least emit a warning if webargs logging is enabled, because this is likely to confuse users.

resolves #518

---

If there's more than we can do, that would be great, but this is all I can think of for that issue. The warning seems reasonable to me (obviously, or I wouldn't submit this PR), but I'm open to discussion.

### DIFF
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -269,6 +269,10 @@ class Parser:
                 error_status_code=error_status_code,
                 error_headers=error_headers,
             )
+            warnings.warn(
+                "_on_validation_error hook did not raise an exception and flow "
+                "of control returned to parse(). You may get unexpected results"
+            )
         return data
 
     def get_default_request(self):


lafrech (approved):
I agree that this should not happen and results are unpredictable. `data` may be `None` or even the result of the deserialization before the second stage of validation if schema loading went fine.

```py
            data = result.data if MARSHMALLOW_VERSION_INFO[0] < 3 else result
            self._validate_arguments(data, validators)
```

Raising would prevent the view function from being entered with bullshit data. Basically, if someone screwed up with `_on_validation_error`, he might think that he's protected whereas in fact he is not, and the function is not entered with data=None (which is likely to error soon) but with data that was loaded but failed validation (which greatly sucks).

This is not a huge flaw in webargs. The dev is to blame. But I don't think it's worth supporting the "`_on_validation_error` does not raise" case.

We can add this warning now and raise in the next major version.

sirosen:
> We can add this warning now and raise in the next major version.

I was about to merge, but I realized that we should have a note of this somewhere.

I could make a `version 7.0` project board ( https://github.com/marshmallow-code/webargs/projects ) and populate it with a note about this, if that sounds good?
I want to clarify before merging and potentially forgetting about this.

lafrech:
I never used the projects feature.

I'd open an issue and add a 7.0 milestone.

sloria:
Would `warnings.warn` be more appropriate here, since this more of a code-level issue rather than something you'd want in application logs?

lafrech:
I guess so: https://docs.python.org/3/howto/logging.html#logging-basic-tutorial

Perhaps I should change this in flask-smorest as well.

In practice, were are those warnings logged?

sloria:
> Warning messages are normally written to sys.stderr, but their disposition can be changed flexibly, from ignoring all warnings to turning them into exceptions.

https://docs.python.org/3/library/warnings.html#module-warnings

lafrech:
Sorry, I could have searched before asking.

https://stackoverflow.com/questions/9595009/warnings-warn-vs-logging-warning

> Python provides a built-in integration between the logging module and the warnings module to let you do this; just call logging.captureWarnings(True) at the start of your script and all warnings emitted by the warnings module will automatically be logged at level WARNING.

https://docs.python.org/3/library/logging.html#logging.captureWarnings

I think I should address this in flask-smorest.

sirosen:
I like the idea of using `warnings` because it's active even if you don't explicitly setup logging for `webargs`.
But in that vein, I would avoid using `DeprecationWarning` because it's suppressed by default, so I'll use the default (`UserWarning`).

Will do that in a few minutes.

> I'd open an issue and add a 7.0 milestone.

I'll do that.

sirosen:
I amended to use `warnings.warn` and kept the text of the warning the same. Should I maybe add a note that we're going to treat this as an error in v7.0 , or just leave it as-is?

sirosen:
I just came back here to check open issues and PRs, and I see this.
I'm going to go ahead and merge. It's effectively the same as when it was first reviewed -- we just swapped `logger.warning` for `warnings.warn`.

lafrech:
Yep. Sorry for not merging this any sooner. I didn't realized you performed the last requested change.

PR #542, opened 2020-09-08
[7.x] Remove support for marshmallow2

### LINKED ISSUE #540: Add support for headers to DjangoParser
```
NotImplementedError: Header parsing not supported by DjangoParser
```

sirosen:
Anywhere that version dispatch was used, switch to the ma3 version of the code. Tests skipped on marshmallow3 are dropped.

Remove dict2schema, since it is now redundant with `Schema.from_dict`.

Update the changelog and docstrings around `unknown=...` to remove note that it only works on marshmallow3. Remove any documentation which notes a marshmallow2 (vs ma3) discrepancy or a way of getting the same behavior out of both versions.

Update readme badge to ma3 only.

tox and azure pipelines config are altered to drop marshmallow2 builds.

Update `setup.py` to require `marshmallow>=3.0.0`

closes #540

---

I did a number of `git grep`s looking for strings like `marshmallow 2` to try to find everything, but there's always a risk that something slipped through. This turned out to be fairly easy, since it's mostly a matter of removing things.


lafrech (approved):
👍

PR #544, opened 2020-09-09
[7.x] Refactor support for 'unknown' with location map

sirosen:
Based on the suggestion in #514, this is my attempt at putting together what I think is ideal behavior for webargs: we have sensible defaults for `unknown` based on locations, but it's easy to opt-out by just setting `unknown=None` or overriding the defaults.
It should be possible to use the `query`, `headers`, and similar locations out of the box without issue. By passing `unknown=EXCLUDE` for these locations, that is achieved.

This introduces a new Parser-class variable, `DEFAULT_UNKNOWN_BY_LOCATION`, which maps location names to
`unknown` values. It populates that and `DEFAULT_UNKNOWN` to pass `EXCLUDE` in the general case and `RAISE` for request bodies.

The new behavior is layered and defined with lower precedence than `Parser.unknown` or the `unknown` parameter on parse calls. In order to implement this and allow users to opt *out* of the behavior (i.e. in order to use schema-defined values), the way these values are set is subtly changed. Instead of having a default value of `None`, parse calls and parser `__init__` have a default of `unknown="_default"`. When that value is detected, the various layers of defaults (DEFAULT_UNKNOWN and DEFAULT_UNKNOWN_BY_LOCATION) are applied. However, if a user passes `None`, that will take effect with the meaning "do not pass a value for `unknown`".

The changelog has been updated significantly in order to handle this and a new section of the advanced usage docs covers the behavior.

If this seems good, I'd like to switch the branches around so that we have a 6.x maintenance branch and 7.0 is in progress in `dev`, as @lafrech suggested. I think we could release this as `7.0.0b1`, if that sounds acceptable. And then other work (like dropping 3.5) can happen during the 7.0 beta period.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,10 +27,11 @@ Refactoring:
 
 Features:
 
-* Add a new ``unknown`` parameter to ``Parser.parse``, ``Parser.use_args``, and
-  ``Parser.use_kwargs``. When set, it will be passed to the ``Schema.load``
-  call. If set to ``None`` (the default), no value is passed, so the schema's
-  ``unknown`` behavior is used.
+* Add ``unknown`` as a parameter to ``Parser.parse``, ``Parser.use_args``, and
+  ``Parser.use_kwargs``, or to parser instantiation. When set, it will be passed
+  to ``Schema.load``. When not set, the value passed will depend on the parser's
+  settings. If set to ``None``, the schema's default behavior will be used (i.e.
+  no value is passed to ``Schema.load``) and parser settings will be ignored.
 
 This allows usages like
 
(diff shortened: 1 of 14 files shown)


lafrech (approved):
LGTM.

Still sounds a bit complicated, but I don't see how we could make things simpler and at least the defaults are sensible.

sirosen:
The only idea I've had for making this simpler which really appeals to me at all is that we could get rid of `Parser.DEFAULT_UNKNOWN`.
That would mean that `DEFAULT_UNKNOWN_BY_LOCATION` is the only way to specify class-level values. Maybe that gets a bit tedious, but it does make the interface smaller and easier to explain.

I also thought about getting rid of `Parser.unknown`, but I'm not sure how I feel about that idea.

I'll leave this open for today, since I don't have time left in the day to handle doing a 7.0.0b1 release and such, as I said I would. But I plan to merge it and do the branch management tomorrow. We can look at removing `DEFAULT_UNKNOWN` after that, if you're interested in that approach.

lafrech:
> The only idea I've had for making this simpler which really appeals to me at all is that we could get rid of Parser.DEFAULT_UNKNOWN.
> That would mean that DEFAULT_UNKNOWN_BY_LOCATION is the only way to specify class-level values. Maybe that gets a bit tedious, but it does make the interface smaller and easier to explain.


Good plan

Doing this before releasing a beta would avoid mentioning `DEFAULT_UNKNOWN` in the changelog, so the user would go straight to `DEFAULT_UNKNOWN_BY_LOCATION` in one step.

sirosen:
I removed `DEFAULT_UNKNOWN` from this branch last night, but I forgot to ask: are you comfortable with me releasing this after merging? I think it's pretty good in this state.

lafrech:
Great. Thanks for taking care of all this.

I suppose you mean release a beta. Sure, go ahead.

There are still a few things in the 7.0 milestone before we ship stable.

sirosen:
Yeah, I meant a beta. Sorry if my phrasing was confusing. I just didn't want to put anything new out -- even a beta -- without giving you a chance to weigh in.
I'll be doing this now.

PR #553, opened 2020-11-02
Drop Python 3.5

### LINKED ISSUE #543: Drop Python 3.5
and update black.

See https://github.com/marshmallow-code/apispec/pull/582

sirosen:
Also update pre-commit config and relevant docs.

There are also a bunch of changes from pyupgrade.

closes #543

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,15 @@
 repos:
 - repo: https://github.com/asottile/pyupgrade
-  rev: v2.7.2
+  rev: v2.7.3
   hooks:
   - id: pyupgrade
-    args: ["--py3-plus", "--keep-mock"]
+    args: ["--py36-plus", "--keep-mock"]
 - repo: https://github.com/psf/black
   rev: 20.8b1
   hooks:
   - id: black
 - repo: https://gitlab.com/pycqa/flake8
-  rev: 3.8.3
+  rev: 3.8.4
   hooks:
   - id: flake8
     additional_dependencies: [flake8-bugbear==20.1.0]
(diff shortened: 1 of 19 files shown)


src/webargs/asyncparser.py
```diff
@@ -1,4 +1,4 @@
-"""Asynchronous request parser. Compatible with Python>=3.5."""
+"""Asynchronous request parser. Compatible with Python>=3.6."""
```
lafrech: This Python version compat comment should be removed.

lafrech (commented):
Thanks for this !

TODO

- Remove comment in async parser
- Update tox.ini: remove py35 (and perhaps update marshmallow>=3.0.0rc2, unrelated but while we're at it)

sirosen:
I knew I must have missed something! 😄 
I'm guessing that comment was added for python2 users, but I wasn't sure.

I've pushed a commit with the changes.

lafrech:
Good. Thanks.

Yes, that comment dates back to when Python2 was still supported and the async parser was in a dedicated py3 directory.

Approved. :+1:

.pre-commit-config.yaml
```diff
@@ -1,15 +1,15 @@
 repos:
 - repo: https://github.com/asottile/pyupgrade
-  rev: v2.7.2
+  rev: v2.7.3
   hooks:
   - id: pyupgrade
-    args: ["--py3-plus", "--keep-mock"]
+    args: ["--py36-plus", "--keep-mock"]
```
sloria, line 6: Do we still need `--keep-mock`?

sirosen: Good catch! I don't think it does anything now that we're not using the backport. I waited to reply until I got to confirm this.
I just opened #556 to clean up.

PR #555, opened 2020-11-05
Switch FalconParser to using 'media' by default

### LINKED ISSUE #253: FalconParser should ideally support falcon's native media decoding
Falcon has a native media handling mechanism which can decode an incoming request body based on the `Content-Type` header and adding the dictionary of resulting key-value pairs as a cached property `req.media`. I've written my own FalconParser subclass that (very naively) uses this, but it seems like something that might be worth supporting out of the box.

```python
    def parse_json(self, req, name, field):
        """
        Pull a JSON body value from the request.
        uses falcon's native req.media
        """
        json_data = self._cache.get("json_data")
        if json_data is None:
            self._cache["json_data"] = json_data = req.media
        return core.get_value(json_data, name, field, allow_many_nested=True)
```

This could probably be improved upon; since the `media` property  is already cached on the request object, we could just access `req.media` directly without caching on the parser. (Not sure if this impacts other things that might use that cache, though; I haven't dug deep enough to fully understand that implication.) Also, since `media` was added in 1.3, if webargs still wanted to support older versions of falcon we could add a check for it and fall back to the existing behavior.

Maybe something like:

```python
    def parse_json(self, req, name, field):
        """Pull a JSON body value from the request.
        .. note::
            The request stream will be read and left at EOF.
        """
        json_data = req.media if hasattr(req, 'media') else self._cache.get("json_data")
        if json_data is None:
            self._cache["json_data"] = json_data = parse_json_body(req)
        return core.get_value(json_data, name, field, allow_many_nested=True)
```

sirosen:
Add a new location to the FalconParser, `media` (handler is `load_media`), and sets it as the default for this parser.

`media` refers to `Request.media` as offered by Falcon, which is a pluggable part of the Falcon request which deserializes data. By
default, it only deserializes JSON paylaods, but you can register additional "handlers", paired with content-types, and it will appropriately map content-type headers to handlers. Falcon docs leave it unclear how exactly the json loading behaves if there is no appropriate payload, so we test these cases explicitly.

New tests for `media` mirror several of the existing JSON tests, but we can skip several cases, like handling of vendor media types (because `media` doesn't do that out of the box). Additionally, we need to avoid expecting webargs-specific error messages in some cases because they won't be returned when error handling is being done by Falcon.

Docs for FalconParser now link to Falcon docs for those who want to customize `media`.

Minor fix slipped in here: `test_parse_json_default` and `test_parse_json_empty` were duplicate tests, so I've removed one.

closes #253

### DIFF
--- a/src/webargs/falconparser.py
+++ b/src/webargs/falconparser.py
@@ -3,6 +3,8 @@
 import falcon
 from falcon.util.uri import parse_query_string
 
+import marshmallow as ma
+
 from webargs import core
 from webargs.multidictproxy import MultiDictProxy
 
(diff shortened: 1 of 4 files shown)


lafrech (approved):
Looks good to me. :+1:

PR #557, opened 2020-11-05
Add support for headers to DjangoParser

### LINKED ISSUE #540: Add support for headers to DjangoParser

sirosen:
Given the choice, use `HttpRequest.headers` rather than `HttpRequest.META`. `headers` was added in Django 2.2 to replace `META`, so it's what modern users should expect in terms of behavior.

The request headers are a case-insensitive mapping in Django, and this should work just fine with schema loading. There's no need to proxy the object since headers can't contain list-like data.

closes #540

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup, find_packages
 
 FRAMEWORKS = [
     "Flask>=0.12.2",
-    "Django>=1.11.16",
+    "Django>=2.2.0",
     "bottle>=0.12.13",
     "tornado>=4.5.2",
     "pyramid>=1.9.1",
--- a/src/webargs/djangoparser.py
+++ b/src/webargs/djangoparser.py
@@ -59,6 +59,9 @@ class DjangoParser(core.Parser):
         return req.COOKIES
 
     def load_headers(self, req, schema):
+        """Return headers from the request."""
+        # Django's HttpRequest.headers is a case-insensitive dict type, but it
+        # isn't a multidict, so this is not proxied
         return req.headers
 
     def load_files(self, req, schema):
--- a/src/webargs/djangoparser.py
+++ b/src/webargs/djangoparser.py
@@ -59,9 +59,7 @@ class DjangoParser(core.Parser):
         return req.COOKIES
 
     def load_headers(self, req, schema):
-        raise NotImplementedError(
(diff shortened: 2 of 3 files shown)


lafrech (commented):
Good.

> Given the choice, use HttpRequest.headers rather than HttpRequest.META. headers was added in Django 2.2 to replace META, so it's what modern users should expect in terms of behavior.

Should we bump min Django version, then?
```py
FRAMEWORKS = [
    "Flask>=0.12.2",
-   "Django>=1.11.16",
+  "Django>=2.2.0",
```

> There's no need to proxy the object since headers can't contain list-like data.

Is this specific to Django? I see headers are MultiDict proxied in Flaskparser.

sirosen:
> Should we bump min Django version, then?

Yeah, good call. I'll do that.

As for the multidict handling...

In the case of Django, headers are exposed as a "CaseInsensitiveDict". If you look through their datastructures, that's not a multidict type.
But in Flask, [`Headers`](https://github.com/pallets/werkzeug/blob/2db33373f4dfadb60dc8cd070cab7e826ccd1ca5/src/werkzeug/datastructures.py#L910) is a multidict implementing `getlist` as a multidict-like method.

Come to think of it, that's really non-obvious, and I think wsgi itself is specified to allow handling of multiple headers with the same name. I'll include a comment in the updated `load_headers` to note why it's not proxied.

lafrech:
Thanks @sirosen for the explanation.

sirosen:
I've pushed an update.

This also made me realize that we don't test against older library versions and I'm not confident we really support these old versions. I'm also going to start work on a branch to figure out the minimal versions of these libs which we actually support and add them to tox. I'll have a PR in a bit, and we can discuss more there in terms of what (if anything) we want in docs and whether or not we want the testing.

lafrech (approved):

lafrech:
> This also made me realize that we don't test against older library versions and I'm not confident we really support these old versions. I'm also going to start work on a branch to figure out the minimal versions of these libs which we actually support and add them to tox. I'll have a PR in a bit, and we can discuss more there in terms of what (if anything) we want in docs and whether or not we want the testing.

Good. Yes, maybe we can test the oldest known version. Let's just try to avoid multiplying the test combinations.

PR #558, opened 2020-11-06
Test against minimum supported framework versions

sirosen:
This adds a tox factor, `mindeps`, which specifies explicitly that we want to test with old versions of the various frameworks.
Right now, it is only specified for our minimum and maximum supported python versions -- meaning 3.6 and 3.9 -- to keep the build matrix a little bit smaller.
These are also added in the azure pipelines config.

This testing is meant to "catch us" if we add a chunk of code to one of the parsers which will break on older framework versions. For example, we might use a request attribute without realizing that it was added in a minor version of a framework. We'd then  break compatibility with older (possibly still supported) framework versions.

This explicitly does not add any public documentation about minimum compatible versions. We don't offer a new guarantee to users and the support we offer is still loose and "best effort". But it does update the versions listed in `setup.py`, so if people are looking there to get a feel for what we support, they'll see something reasonable.

flask and aiohttp both failed on these builds. flask because it didn't start pinning the version of werkzeug until 0.12.5, and aiohttp because versions prior to 3.0.8 aren't py3.9-compatible. So these have been updated.

---

`webapp2` isn't actually supported and I think it should be removed from the list in `setup.py`, but I refrained from doing that for now.

I considered only testing on the oldest supported python (3.6), and I wouldn't mind doing it that way.
I also considered maybe adding a chunk of doc about supported/compatible versions, but nobody is asking us for a written policy. So I don't see a need for one at present.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,7 +31,6 @@ jobs:
       - py37
       - py38
       - py39
-      - py39-mindeps
       - py39-marshmallowdev
       - docs
     os: linux
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,9 +27,11 @@ jobs:
     toxenvs:
       - lint
       - py36
+      - py36-mindeps
       - py37
       - py38
       - py39
+      - py39-mindeps
       - py39-marshmallowdev
       - docs
     os: linux
--- a/setup.py
+++ b/setup.py
@@ -2,14 +2,14 @@ import re
 from setuptools import setup, find_packages
 
 FRAMEWORKS = [
-    "Flask>=0.12.2",
+    "Flask>=0.12.5",
     "Django>=2.2.0",
     "bottle>=0.12.13",
     "tornado>=4.5.2",
     "pyramid>=1.9.1",
     "webapp2>=3.0.0b1",
     "falcon>=2.0.0",
-    "aiohttp>=3.0.0",
+    "aiohttp>=3.0.8",
 ]
 EXTRAS_REQUIRE = {
     "frameworks": FRAMEWORKS,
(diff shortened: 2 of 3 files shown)


lafrech (commented):
Good.

While we're at it, and since we're about to release a new minor version, I'd be tempted to bump min deps to major versions: Flask 1.0.0 has been out for a while, now, and the move is easy to make from 0.x, I believe. Also, I don't think the 0.x branch is maintained. But this it not mandatory at all. We don't have to force the user to move and AFAIK, supporting 0.12.5 costs us zero.

> webapp2 isn't actually supported

Right, looks like it is not maintained. https://pypi.org/project/webapp2/#history

I don't mind dropping it. Can be done in another PR.

> aiohttp [...] versions prior to 3.0.8 aren't py3.9-compatible

Perhaps a case for testing min deps only on oldest Python version. No need to push users of old Py versions to new dep versions just because we want to test on new Py versions. In this case, it is just 3.0.0 -> 3.0.8 but if Py 3.9 was only supported on e.g. 4.x, we'd have to bump min dep to 4.0 which would be a pity.

Only testing on old Py version solves this (mostly theoretical for now) issue.

> I also considered maybe adding a chunk of doc about supported/compatible versions, but nobody is asking us for a written policy. So I don't see a need for one at present.

We can stick to the list in setup.py for now.

sirosen:
I'm not particularly picky about exactly versions we put in `setup.py` since we're never interfering with what versions actually get installed for users. I'll leave the flask one alone for now, but I'd also have no objection to changing it.

Why don't I cut this down to only running the "mindeps" test on our oldest python version? I don't have a strong feeling about this, and even if the issue you bring up is only theoretical, accounting for it in the test config now potentially saves us from having to deal with it or make a decision later on.

lafrech:
Approved.

PR #559, opened 2020-11-09
Fix async tests to not be order-sensitive

### LINKED ISSUE #465: async tests are execution order dependant
Moving aiohttp tests from test_py3 into root test dir was forgotten in @hugovk's #454.

When I do this I get an error in a tornado test that should be unrelated.

No time to investigate right now.

sirosen:
Fixes #465

Primarily, this is about fixing #465, but there was a little bit of related cleanup in here as well.

The Tornado tests directly constructed a `Future` object. Modern tornado docs state that's an alias for `asyncio.Future`. This is the part that was flakey: a `Future` can only be constructed when there is a current/active event loop. _For some reason which I was not able to fully determine_, when `tests/test_py3/test_aiohttpparser_async_functions.py` runs before the tornado tests, it leaves us without an event loop, and then creating a `Future` fails.
I was able to fix this by trying to change that test module to explicitly create and close event loops instead of using the pytest-aiohttp `loop` fixture, as in

```python
@pytest.fixture()
def testapp(app):
    loop = asyncio.new_event_loop()
    yield webtest_aiohttp.TestApp(app, loop=loop)
    loop.close()
```

but I'll be up-front in saying that I don't understand why this worked whereas the `loop` fixture didn't.

Rather than taking that fix, I went in a different direction: if pytest-aiohttp is going to provide a working event loop as a fixture, I'd like to use that for any tests which need an event loop. So in the tornado tests, I've started using `loop`, and I've added it to the aiohttp tests to replace usage of `asyncio.get_event_loop`.

After moving the aiohttp tests up a directory, I reviewed the two aiohttp test modules. I was unable to determine anything being tested in `test_aiohttpparser_async_functions.py` which was not covered by `test_aiohttpparser.py`.
I've removed the `async_functions` tests entirely as redundant.
However, I tagged in @sloria for review because he's the primary/original author of those tests -- maybe there's a subtlety there which I'm missing?

As a small fix, I've removed use of `asyncio.coroutine` in favor of `async def` since `asyncio.coroutine` is being removed in py3.10. So it just behooves us to do this cleanup.


lafrech:
Great. Thanks for this. I had been digging into it but to no conclusion, as documented in #465.

I'm not qualified to tell if this is "the right fix". Maybe "the right fix" would be to address https://github.com/sloria/webtest-aiohttp/issues/6. Or maybe this fix makes it irrelevant.

In any case, I'm happy with this.

lafrech:
I don't think we have to wait on @sloria for this. Those changes sound reasonable.

We could merge this and release.

In fact we could release 7.0.0 (not beta). We can also drop webapp2 before that. Fixing #563 could be a breaking change but we can wait for a next major release as it will take a bit of time.

sirosen:
Yeah, I was considering asking if you'd agree that it's fine to just move forward, but I didn't want to nag.

> In fact we could release 7.0.0 (not beta).

Yeah, I've been uncertain about when to make the call on this. I'd like to do 7.0.0b2 just because I've already queued that up (and because it gives us a moment to think about whether or not we think we can sneak in any other issues before finalizing 7.0). But starting tomorrow I'll think about releasing 7.0.

lafrech:
Alright. Not off, just swamped with other stuff. I just sent a PR to drop webapp2.

To me that's enough to ship a 7.0.0 but if you want to cram other changes in there, no pb.

PR #562, opened 2020-11-18
Update changelog for v7.0.0b2

sirosen:
I was thinking that I would like to release v7.0.0b2 soon and I noticed that I haven't been putting in changelog entries the way I should.

This just gets the changelog up to date, and I'm happy to release 7.0.0b2 whenever this merges if that sounds good.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,9 +8,9 @@ Features:
 
 * `DjangoParser` now supports the `headers` location. (:issue:`540`)
 
-* `FalconParser` now defaults to the new `media` location, which uses
+* `FalconParser` now supports a new `media` location, which uses
   Falcon's `media` decoding. (:issue:`253`)
-  
+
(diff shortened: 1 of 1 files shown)


lafrech (commented):
Sorry, I totally forgot about this.

Good.

Just a nitpick:

This part sounds redundant as it is repeated twice. 

> FalconParser` now defaults to the new `media` location

I think I understand the logic:

- feature: new media parser
- change: media is default

If so, we could keep it in two places as

- feature: new media parser (don't mention default)
- change: media is default

or merge it into:

- feature: new media parser, which becomes default.

I don't mind, both are totally valid, no need to overthink.

Feel free to merge after or even without rewording.

sirosen:
No worries! I figured you might be off for a few weeks.

I'll remove mention of "default" from the "feature" section and merge. I wanted to highlight it as a change since it could be backwards-incompatible for some people (at least in theory).

PR #565, opened 2020-12-02
Drop webapp2 support

lafrech:
Looks like webapp2 is unmaintained.

https://pypi.org/project/webapp2/#history


sirosen:
Thanks for doing this. I can't find any other references to webapp2 with a `git grep`, so I think you got them all. I'll merge.

PR #567, opened 2020-12-04
Add type annotations to Parser and fields

sirosen:
The type annotations on the parser base class are mostly ported over from AsyncParser with some workarounds and other trickery (documented in comments and commit messages) based on two `mypy` issues I ran into and one bit of `async` vs non-`async` conflict.

The conflicts between `Parser` and `AsyncParser` make me want to try more radical refactoring at some point. The fact that we have async methods whose names match their synchronous counterparts is problematic. Maybe some shared base class for both of them could get us out of their direct inheritance relationship? It needs more thought but not for inclusion here.

---

I think this is good to go and is ready for review, but I want to actually try running `mypy` against some projects which use `webargs` to see if I can find issues that way. So if it's approved, I might hesitate to merge until I can do that.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,9 @@ Changes:
 
 * *Backwards-incompatible*: Drop support for webapp2 (:pr:`565`).
 
+* Add type annotations to `Parser` class, `DelimitedList`, and
+  `DelimitedTuple`. (:issue:`566`)
+
 7.0.0b2 (2020-12-01)
 ********************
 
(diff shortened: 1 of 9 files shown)


sirosen:
I just made a small addition to this PR to add the `py.typed` file ( https://www.python.org/dev/peps/pep-0561/#typed-package ). I was trying to check some code which depended on webargs against this and things were coming up empty in `reveal_type`. It took a moment to figure out that this was missing.

I don't know that this needs to be explicitly called out in the changelog, so I've left it out for now. Arguably it's implicit in "added annotations for ..."

lafrech (approved):
LGTM

I didn't do a thorough line by line review. Besides, I don't know much about typing.

But I agree it should be fine once checked with your codebase.

sirosen:
I'm going to move forward with the 7.0 release, since you've given the signal in a previous comment that we're ready.
This was the last thing I wanted to get added.

PR #568, opened 2020-12-10
Add section to upgrading doc for 7.0

sirosen:
After 7.0, I looked over things to see if there are/were doc improvements we need to make. This was the one thing that jumped out -- the upgrading doc should have a new section.

Focus on explaining the changes to `unknown` and clarifying for anyone on 6.x that to use their schema's settings, they should set `unknown=None`.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -127,6 +127,7 @@ When you need more flexibility in defining input schemas, you can pass a marshma
         username = args["username"]
         # ...
 
+.. _advanced_setting_unknown:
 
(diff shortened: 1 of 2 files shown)


lafrech (approved):
Good.

This is obvious:

    # assuming you have a schema named MySchema

but you can leave it, no no pb.

sirosen:
Yeah, I probably could have skipped that note, but sometimes people ask questions that I think are obvious.
So sometimes I think I just can't tell. 🤷

PR #570, opened 2020-12-11
Fix issues with type annotations and pass through kwargs to _serialize in delimited fields

### LINKED ISSUE #569: Incorrect method signature
Definition of `_serialize` in class `DelimitedFieldMixin` is incompatible with definition of this method in marshmallow classes `FieldABC`, `Field`, `List`, etc.
https://github.com/marshmallow-code/webargs/blob/91c4408756c5c9f911f1b93d79591290d802bce1/src/webargs/fields.py#L59

For instance https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/base.py#L25, so `**kwargs` are absent.

sirosen:
This resolves #569 

In the particular case cited there, `DelimitedList._serialize` and `DelimitedTuple._serialize` weren't taking additional keyword arguments even though that's required to be compatible with the function signature in `Field._serialize`.
In addition to taking the args, use the same behavior as `marshmallow.fields.List` and pass the keyword args through to the inner field's `_serialize` method.

I also found several much more minor issues when trying to run `mypy` from an environment which had webargs installed, and I've resolved all of them (sometimes by removing annotations which we can come back to later).

To ensure that this kind of flaw can't creep back in, I've added `tox -e mypy` as a separate test env which runs mypy, and added config to run that test in azure pipelines.

I intend to release this right away as 7.0.1 since the type signature issue could be stopping people from upgrading.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+7.0.1 (Unreleased)
+******************
+
+Bug fixes:
+
+* Fix `DelimitedList` and `DelimitedTuple` to pass additional keyword arguments
+  through their `_serialize` methods to the child fields and fix type checking
+  on these classes. (:issue:`569`)
+  Thanks to :user:`decaz` for reporting.
+
 7.0.0 (2020-12-10)
 ******************
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,6 +26,7 @@ jobs:
   parameters:
     toxenvs:
       - lint
+      - mypy
       - py36
       - py36-mindeps
       - py37
--- a/src/webargs/aiohttpparser.py
+++ b/src/webargs/aiohttpparser.py
@@ -25,7 +25,6 @@ Example: ::
 import typing
 
 from aiohttp import web
-from aiohttp.web import Request
 from aiohttp import web_exceptions
 from marshmallow import Schema, ValidationError, RAISE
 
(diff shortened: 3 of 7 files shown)


lafrech (approved):
> I intend to release this right away as 7.0.1 since the type signature issue could be stopping people from upgrading.

Agreed. You may merge. It's never too late to come back to it later on.

It's a bit unclear to me why it fails. And if this trick is needed, it should be needed on our other projects as well, shouldn't it?

I remember having an issue where mypy would pass on my computer and fail in CI. Can't remember exactly why. Perhaps unrelated.

In any case, the changes here can be reworked in a non-breaking manner if we ever sort this out with a better solution than adding another test env, so let's release the fix ASAP.

sirosen:
> It's a bit unclear to me why it fails. And if this trick is needed, it should be needed on our other projects as well, shouldn't it?

Yes, I'm a bit concerned about this as well. It might be a gap in the testing/linting for other projects, but obviously it only matters if someone makes a mistake. So perhaps we've gotten lucky / been sufficiently careful where it matters.

One part that I can explain for sure is that not having `marshmallow` installed into the environment where `mypy` runs means that `marshmallow`'s types and signatures (like `FieldABC`) aren't available. So the error in which `DelimitedList._serialize` doesn't match the signature for `FieldABC._serialize` can't be detected by `mypy` in that context.

I think there's another issue as well, but I'm not sure exactly what the cause is. Running `mypy src/` found inconsistencies between `Parser` and `AIOHTTPParser` which weren't found before. I'm going to try to figure that out.

lafrech:
If it's only about having having marshmallow installed, would it work by just removing `skip_install ` in lint test env?

sirosen:
`pre-commit`, like tox, manages virtualenvs, so we've got layers of virtualenv management going on. We could add `marshmallow` to the `additional_dependencies` list for `mypy` in pre-commit config, which would get it into the right place when `mypy` runs.

That definitely improves things, but I don't think it would solve the part of this that I still don't understand. I'm hopefully going to get some time this evening to experiment and try to work out what works.

PR #571, opened 2020-12-16
Replace mypy tox target with fixed precommit conf

sirosen:
In 7.0.1 I introduced a new tox target because I wasn't sure about what had gone wrong with `mypy` and didn't want to take risks.
It turns out, that was overkill. After testing, I'm quite confident that the _only_ problem was that `marshmallow` was not installed into the `mypy` environment. Detail (per the commit message) below.

You can test this by checking out `7.0.0` and trying this config against it -- you'll get pre-commit errors from the few bad signatures which were present in that version.

cc @sloria Even though you haven't been involved, I thought I ought to give you a heads up in case this impacts other projects too.

---

The issues with undiscovered type errors actually derived from the fact that `marshmallow` was not available in the `mypy` environment.
As a result, values like `missing` could not be resolved and were typed as `Any`. That, in turn, leads to "strangely incorrect" signatures like `Union[Dict, Any]` (a.k.a. `Any`), which then can result in type errors that ought to be identified being missed.

Once `marshmallow` is installed, mypy will find types like `Union[Dict, _Missing]`, and can determine that they do (or do not) match as appropriate.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,3 +25,4 @@ repos:
   - id: mypy
     language_version: python3
     files: ^src/webargs/
+    additional_dependencies: ["marshmallow>=3,<4"]
--- a/tox.ini
+++ b/tox.ini
@@ -30,6 +30,7 @@ commands = pre-commit run --all-files
 # issues in which `mypy` running on every file standalone won't catch things
 [testenv:mypy]
 deps = mypy
+extras = frameworks
 commands = mypy src/
 
 [testenv:docs]


sloria:
Here's an alternative approach I've used: https://github.com/sloria/environs/blob/91d60fb21719650e7bfce104f7ad06fc0f632f63/.pre-commit-config.yaml#L22-L30 . Basically, run `mypy` in the "development" virtualenv so that it can check against all dependencies' types

sirosen:
Ah, and then you have `tox` configured to install things into the pre-commit env. That's clever.

The only thing I don't like about that is that `pre-commit` could give a false-positive when I commit (if the development env isn't active when `git commit` runs). But CI would catch it, so it's not that bad.

I think I prefer this at the moment, but if you think the other way is significantly better, I'm not opposed.

sloria:
I've no strong preference either way. But with the PR as is, won't the types for all the web frameworks still be missing?

sirosen:
Coming back to this, the gap you note in the setup I put forth is real -- it doesn't catch type errors which can only be seen with frameworks installed. But, and this is personal taste, I don't really like the idea of `pre-commit` being dependent on the surrounding environment if it can be avoided.

So I refined my branch to keep both: the pre-commit improvement to install `marshmallow` and the `tox` + CI setup to run `mypy` with everything installed, now including frameworks too.

I can definitely see an argument for just going with the system invocation of `mypy` in pre-commit, but this is what I prefer.

lafrech:
I don't really care either way.

The 'run mypy in the "development" virtualenv' solution only has one shortcoming: it might not detect errors if the dev env is not installed.

The other solution requires making sure the list of deps is up-to-date in tox.ini, and if it is not it can go unnoticed. But this list shouldn't evolve so not really an issue.

@sirosen I'll follow your preference.

sirosen:
I'm inclined towards what I have written -- I think using the `frameworks` extra in `tox` should be easy enough to maintain.

I'm happy to follow anyone's strong feelings about what we should do, but if you're satisfied with what I have in this PR, please merge! 😄

Issue #563, opened 2020-11-20
`is_multiple` vs custom fields


ThiefMaster:
Right now I don't get multiple values in a custom field that does not inherit from the `List` field because `is_multiple` only checks for this.

And for some cases rewriting the field to be nested in an actual List field is not feasible, for example a `SQLAlchemyModelList`-like field where sending individual queries would be inefficient compared to one query with an SQL `IN` filter.

I propose adding a `MultiValueField` which would simply inherit from Field and do nothing itself, and instead of checking for isinstance List check for that one in `is_multiple`. `List` would then of course inherit from that instead of `Field`.

Alternatively, check for something like `isinstance(field, List) or getattr(field, 'is_multiple', False)`.

I can send a PR if there's interest in this.

sirosen:
I admit that I might not be well enough educated about marshmallow-sqlalchemy (I've never used it) to be able to make a good judgement about that part of this issue. But clearly, `is_multiple` isn't very extensible right now, and that's a problem. (Also, I think it doesn't handle `Tuple`? So that's a bug.)

If we put `MultiValueField` into `marshmallow`, that might get confusing with `Nested(..., many=True)` as a possibility. I think `is_multiple` makes more sense as a characteristic of a field instance than a field class -- that makes it possible (at least in theory) to have a field whose instances vary with respect to this.

As a result, I think the `getattr` solution is the right one. `is_multiple` seems like a reasonable name for the attribute to me. @lafrech, do you maybe have a different opinion about this, or any thoughts about naming?

sirosen:
Although this was closed in #584, I'm working on a follow-up change that should document the interface for this and make it even easier to customize. I plan to put that together potentially for v7.1.0 .
I'll try to remember to follow up here when we release the new version, but I may forget.

sirosen:
I just pushed out v8.0 , which turned out to be the first version supporting `is_multiple` on fields.

In contrast with the last couple of major releases in webargs, this one has fairly modest changes, so it should be fairly easy to update.

lafrech:
Great. Thanks.

lafrech:
I did a few CHANGELOG updates. I don't think it was worth a PR so I pushed to dev.

lafrech:
Also we might want to add a paragraph in the upgrading section of the docs.

https://webargs.readthedocs.io/en/latest/upgrading.html

PR #583, opened 2021-02-01
Add `Parser.pre_load` as an extension point

### LINKED ISSUE #234: Automatically trim leading/trailing whitespace from argument values
Does webargs provide any clean way to do this? I guess leading/trailing whitespace are almost never something you want (especially when having required fields that must not be empty)...

sirosen:
This is a potential solution to #234 which doesn't put too much maintenance burden on us.
I'm happy to go down other paths (e.g. allow usage as a decorator, like error handlers), but this struck me as the solution with the absolute least amount of new code needed to solve the problem.

As an added aside, I'd like to handle this and take a look at #563 in 7.x to see what we can/should do, and then we can look at anything we want or need for 8.0. That way, if we'd like to take a route on these issues which requires a major version bump, we'll keep churn to a minimum.

---

`Parser.pre_load` is a "do nothing" method out of the box, but lets users subclass and override with custom implementations for their needs. This resolves #234 by allowing a custom parser to strip whitespace -- or perform any other desirable action -- after location loading and before schema loading.

By giving the pre_load method an interface which takes the schema and location, users can extend this in a myriad of ways.

In an explicit nod towards #234, an example is provided which strips whitespace from certain locations. Furthermore, since satisfying this use case is the purpose of this PR, I've included a somewhat unusual test which copies the doc example code and ensures it does the right thing in a couple of basic scenarios.

Defining this interface as a method for subclasses to override keeps the changes to `webargs` minimal (vs. supporting another hook with a decorator, like error handlers and location loaders).

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,4 +1,5 @@
 import datetime
+import typing
 from unittest import mock
 
 import pytest
@@ -36,6 +37,9 @@ class MockRequestParser(Parser):
 
     def load_querystring(self, req, schema):
         return self._makeproxy(req.query, schema)
+
+    def load_form(self, req, schema):
+        return MultiDictProxy(req.form, schema)
 
     def load_json(self, req, schema):
         return req.json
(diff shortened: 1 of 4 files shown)


sirosen:
@lafrech, can you take a look at this for 8.0 and tell me if you think it's okay to add this (or something similar) or if you think we should do something different? My goal is to have a solution for #234 .

I looked at making `Parser._load_location_data` into a public interface so users could just override and wrap that, but it turned out messier and harder to document than this PR.

tests/test_core.py
```diff
@@ -1134,3 +1138,83 @@ class CustomParser(Parser):
...
+    schema = Schema.from_dict(
+        {"ids": fields.List(fields.Int), "values": fields.List(fields.Str)}
+    )
+    for loc in ("query", "form"):
+        ret = parser.parse(schema, web_request, location=loc)
+        assert ret == {"ids": [1, 3, 4], "values": ["foo", "bar"]}
+
+    # for this case, json loading should also work because `int("  1")` is
+    # allowed, but note that the string list arg won't be adjusted with
+    # whitespace even though that's "part of the argument"
```
lafrech: I found that comment unclear.

(Not sure this test is necessary but it doesn't hurt to keep it to check the doc example works.)

lafrech (approved):
Sorry, I totally forgot about this.

Absolutely fine for me. Should make it into 8.0.

tests/test_core.py
sirosen: I've amended to try to make it clearer. On review, I also was baffled: "what was I talking about?!" Let me know if the fixed comment seems better.

Also, I agree that the test isn't strictly necessary. But I was already doing the work to validate that the example works, so it seemed like I might as well turn it into a test.

lafrech:
Good to go.

Thanks.

sirosen:
Awesome!

I'll wait for us to sort out other changes and then I'll rebase and merge this. I'm hoping we can fixup the changelog in those, so that this can just have a clean addition to the changelog.

sirosen:
@lafrech, this is the last thing slated for 8.0, I think, but I'm not in a hurry to release.

I think I'll take a look at our open issues tonight or tomorrow and do a release if I don't see anything which we want to include.
Let me know if you want me to hold off or consider including anything.

lafrech:
Great. Feel free to release.

Thanks.

PR #584, opened 2021-02-02
Add support for the `is_multiple` field attribute

### LINKED ISSUE #563: `is_multiple` vs custom fields

sirosen:
This resolves #563

Detection of fields as `is_multiple` for multi-dict unpacking now allows custom fields to use an attribute `is_multiple=T/F` to control the behavior directly. This allows for a subclass of `List` to set `is_multiple=False` (done by `DelimitedList`) and allows non-`List` subclasses to be detected as `multiple`.

Because `is_multiple` is an attribute, it can be set on the field class or on an instance.

A noted TODO comment questions whether or not `ma.fields.Tuple` should be added to the auto-detected fields for `is_multiple`. I wasn't certain if we should add it or not, but I think so? I'll open a follow-up issue to decide on it for 8.0 -- I don't think we should do it in 7.x since 8.0 is on the horizon.

Tests cover
- a field with is_multiple=True
- a field with is_multiple=False
- a field where is_multiple is not set
- a List(Str()) field

and run against the collection of multidict implementations used to test MultiDictProxy.

### DIFF
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -35,8 +35,8 @@ T = typing.TypeVar("T")
 # a set of fields which are known to satisfy the `is_multiple` criteria, but
 # which come from marshmallow and therefore don't know about webargs (and
 # do not set `is_multiple=True`)
-# TODO: should `ma.fields.Tuple` be added here?
-KNOWN_MULTI_FIELDS: typing.Tuple[typing.Type, ...] = (ma.fields.List,)
+# TODO: `ma.fields.Tuple` should be added here in v8.0
+KNOWN_MULTI_FIELDS: typing.List[typing.Type] = [ma.fields.List]
 
 
 # a value used as the default for arguments, so that when `None` is passed, it
@@ -68,7 +68,7 @@ def is_multiple(field: ma.fields.Field) -> bool:
     is_multiple_attr = getattr(field, "is_multiple", None)
     if is_multiple_attr is not None:
         return is_multiple_attr
-    return isinstance(field, KNOWN_MULTI_FIELDS)
+    return isinstance(field, tuple(KNOWN_MULTI_FIELDS))
 
 
 def get_mimetype(content_type: str) -> str:
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -8,8 +8,6 @@ import marshmallow as ma
 from marshmallow import ValidationError
 from marshmallow.utils import missing
 
-from webargs.fields import DelimitedList
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


src/webargs/core.py
```diff
@@ -34,6 +32,12 @@
 # generic type var with no particular meaning
 T = typing.TypeVar("T")
 
+# a set of fields which are known to satisfy the `is_multiple` criteria, but
+# which come from marshmallow and therefore don't know about webargs (and
+# do not set `is_multiple=True`)
+# TODO: should `ma.fields.Tuple` be added here?
+KNOWN_MULTI_FIELDS: typing.Tuple[typing.Type, ...] = (ma.fields.List,)
```
lafrech: Should this be mutable to allow users to add field classes here?

lafrech (approved):
Sorry for the delay.

LGTM.

I'd be tempted to consider #585 a bugfix, but if 8.0 is on its way, this can wait.

The feature might deserve to be documented but this is not a blocker. We can release this as is in 7.x for curious users and advertise it in the docs later.

src/webargs/core.py
sirosen: I didn't really think about this var as part of the public interface. I'm not against it, I just didn't think of that at all! 😄

If we start going down the path of it being customizable I might want to move it onto the parser so that it can be controlled by subclassing. Right now pretty much all configurable behavior can be set on a parser. Should I take a look at doing that?

sirosen:
> I'd be tempted to consider #585 a bugfix, but if 8.0 is on its way, this can wait.

Yeah, if we were going to take a long time to do 8.0, I would maybe be more inclined to put it in as a bugfix; maybe included here. But I just want to take care of whatever can be done without a major version bump and then go ahead and do 8.0 with the changes we've planned.

I could add something in the advanced doc on this... Let's circle back on that later though? I think the changelog is doc enough for the moment.

src/webargs/core.py
lafrech: Sure. Good point.

But no big deal if we keep it as a global. It's not like someone will be instantiating several parsers with different `KNOWN_MULTi_FIELD` values.

sirosen: I think you're right that multiple parsers with different values is an unlikely/strange use-case.

It's getting late in my local time (US Eastern), so I'll come back to this tomorrow. But right now I'm inclined to change it to be a list and merge the result. At least that way, it's not actively stopping anyone from modifying it.

lafrech: Yeah, it is not a really friendly API, but it doesn't get in the way of advanced users.

sirosen:
I converted `KNOWN_MULTI_FIELDS` to a list ( 7f93baa ) and merged on the basis of the prior approval.
I'll take a look at making that configurable and documented as a parser attribute.

PR #592, opened 2021-03-08
Make KNOWN_MULTI_FIELDS configurable and add doc

sirosen:
Rather than a module-level constant, move ``KNOWN_MULTI_FIELDS`` to be a parser attribute. This will allow users to override or modify it. However, moving it creates a problem: whereas previously `is_multiple` was just a function which MultiDictProxy could safely call, now it's a method and requires a Parser instance.
(NB: `is_multiple` actually moved from Parser to MultiDictProxy, but my first draft put it on the parser.)

To handle this, change the relationship between the multidictproxy and core modules. `core` now requires `multidictproxy` and the base `Parser` provides a private method for building a MultiDictProxy object.

MultiDictProxy doesn't depend on parsers (tried that, didn't like it). Instead, a MultiDictProxy can be given a `known_multi_fields` argument when created. The `_makeproxy` helper on the Parser class just makes sure to pass `known_multi_fields=self.KNOWN_MULTI_FIELDS`, and everything falls into place from there. Numerous tweaks to parser classes use `_makeproxy` instead of directly instantiating a MultiDictProxy, but that's the bulk of the changeset.

New docs in a section towards the end of the advanced doc explains how multi field detection works and explains how you can set `is_multiple = True` or add a field class to `Parser.KNOWN_MULTI_FIELDS`.

---

If we like this, I think I'll release v7.1 when it merges. If we don't, let's figure out what kind of doc & interface we want instead.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -493,6 +493,91 @@ For example, you might implement JSON PATCH according to `RFC 6902 <https://tool
         """
         # ...
 
+Multi-Field Detection
+---------------------
+
+If a ``List`` field is used to parse data from a location like query parameters --
+where one or multiple values can be passed for a single parameter name -- then
+webargs will automatically treat that field as a list and parse multiple values
+if present.
+
+To implement this behavior, webargs will examine schemas for ``marshmallow.fields.List``
+fields. ``List`` fields get unpacked to list values when data is loaded, and
+other fields do not. This also applies fields which inherit from ``List``.
+
+.. note::
+
+    In webargs v8, ``Tuple`` will be treated this way as well, in addition to ``List``.
+
+What if you have a list which should be treated as a "multi-field" but which
+does not inherit from ``List``? webargs offers two solutions.
+You can add the custom attribute `is_multiple=True` to your field or you
+can add your class to your parser's list of `KNOWN_MULTI_FIELDS`.
+
+First, let's define a "multiplexing field" which takes a string or list of
+strings to serve as an example:
+
+.. code-block:: python
+
+    # a custom field class which can accept values like List(String()) or String()
+    str_instance = fields.String()
+
+
+    class CustomMultiplexingField(fields.Field):
+        def _deserialize(self, value, attr, data, **kwargs):
+            if isinstance(value, str):
+                return str_instance.deserialize(value, **kwargs)
+            return [str_instance.deserialize(v, **kwargs) for v in value]
+
+        def _serialize(self, value, attr, data, **kwargs):
+            if isinstance(value, str):
+                return str_instance._serialize(value, **kwargs)
+            return [str_instance._serialize(v, **kwargs) for v in value]
+
+
+If you control the definition of ``CustomMultiplexingField``, you can just add
+``is_multiple=True`` to it:
+
+.. code-block:: python
+
+    # option 1: define the field with is_multiple = True
+    from webargs.flaskparser import parser
+
+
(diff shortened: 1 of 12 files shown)


docs/advanced.rst
```diff
@@ -493,6 +493,91 @@ For example, you might implement JSON PATCH according to `RFC 6902 <https://tool
...
+---------------------
+
+If a ``List`` field is used to parse data from a location like query parameters --
+where one or multiple values can be passed for a single parameter name -- then
+webargs will automatically treat that field as a list and parse multiple values
+if present.
+
+To implement this behavior, webargs will examine schemas for ``marshmallow.fields.List``
+fields. ``List`` fields get unpacked to list values when data is loaded, and
+other fields do not. This also applies fields which inherit from ``List``.
```
lafrech: applies *to* fields

docs/advanced.rst
```diff
@@ -493,6 +493,91 @@ For example, you might implement JSON PATCH according to `RFC 6902 <https://tool
...
+First, let's define a "multiplexing field" which takes a string or list of
+strings to serve as an example:
+
+.. code-block:: python
+
+    # a custom field class which can accept values like List(String()) or String()
+    str_instance = fields.String()
+
+
+    class CustomMultiplexingField(fields.Field):
```
lafrech: Why the instance? Couldn't this inherit from String using super to access String methods?

lafrech (approved):
LGTM. Thanks.

I've been wondering if this should be a major release.

Overrides in parsers using `MultiDictProxy` (`load_querystring`,...) will still work like they used to, just ignoring the `is_multiple` rework.

Should removing `is_multiple` from `core.__all__` be considered a breaking change?

docs/advanced.rst
sirosen: Yes, it could/should!
I had just imitated the implementation of `List` (which has a nested field instance) without thinking too hard.

I'll simplify it.

sirosen:
> Should removing `is_multiple` from `core.__all__` be considered a breaking change?

I hadn't thought of it as public, but it's certainly possible that someone is importing `is_multiple` and using it.
Which leads to...

> I've been wondering if this should be a major release.

If so (mostly, if we're concerned that removing `is_multiple` could break someone), but you're happy with it as-is, let's mark it as part of 8.0 .
The existing work in the dev branch (changing `is_multiple` behavior to respect the `is_multiple` attribute) could still go out as 7.1 , so we should refrain from merging.

Would you like me to release 7.1 with just what we have today? I'm happy to do so.

lafrech:
I don't mind if you ship 7.1 with current dev, then this PR later in 8.0.

But the whole `is_multiple` change (dev + this PR) could be shipped as 8.0. I mean it is simpler to make the whole thing a single version jump, be it major, with all the changes at once without having to document intermediate steps. The `Tuple` change would be part of it as well.

It's not like someone really needs this feature but can't move to 8.0. The only change with 8.0 is related to the feature anyway.

sirosen:
Yeah, good point. Why don't we commit to `dev` being the working branch for 8.0 now? That way it's simpler for us and we get to just document the `is_multiple` change for 8.0.

I'm doing cleanup right now to this branch; hopefully done in a few minutes.

lafrech:
> Why don't we commit to dev being the working branch for 8.0 now? That way it's simpler for us and we get to just document the is_multiple change for 8.0.

Exactly.

PR approved.

PR #594, opened 2021-03-12
Add Tuple to KNOWN_MULTI_FIELDS

### LINKED ISSUE #585: Add marshmallow.fields.Tuple to detected `is_multiple` fields
I noticed this while working on #584 . Currently, `is_multiple(ma.fields.List(...)) == True`, but `is_multiple(ma.fields.Tuple(...)) == False`.
We should add `Tuple` so that `is_multiple` returns true.

It is possible that users have subclassed `fields.Tuple` or done something else which could make that a breaking change. I don't think that's very likely, but we are already queuing things up for 8.0 . I'm going to put this in the 8.0 milestone. We could probably do it sooner if there's a reason to do so.

I don't think using `Tuple` for a query param is a good idea. Supporting `...?foo=1&foo=2` but *not* `...?foo=1` or `...?foo=1&foo=2&foo=3` is unusual and therefore likely to be confusing. I'd advise webargs users not to do that. However! It's surprising that `Tuple` is not treated as `is_multiple=True`, and I'm just aiming for the least surprising behavior for webargs here.

sirosen:
Add Tuple to known multi fields both in the parser default and the default arg for `MultiDictProxy`.

Update changelog as appropriate and fix some single-backticks which should be double backticks. Mark the unreleased changes as 8.0.0 .

resolves #585

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -35,7 +35,7 @@ class MockRequestParser(Parser):
     """A minimal parser implementation that parses mock requests."""
 
     def load_querystring(self, req, schema):
-        return MultiDictProxy(req.query, schema)
+        return self._makeproxy(req.query, schema)
 
     def load_json(self, req, schema):
         return req.json
@@ -1040,6 +1040,7 @@ def test_type_conversion_with_multiple_required(web_request, parser):
         "is_multiple_false",
         "is_multiple_notset",
         "list_field",
+        "tuple_field",
         "added_to_known",
     ],
 )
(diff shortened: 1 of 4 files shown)


tests/test_core.py
```diff
@@ -1103,13 +1104,20 @@ class CustomNonMultipleField(CustomMultiplexingField):
...
+        assert result["foos"] in (("a", "b"), ("b", "a"))
+    # case 6: the field is custom, but added to the known fields of the parser
     elif setting == "added_to_known":
         # if it's included in the known multifields and is_multiple is not set, behave
         # like is_multiple=True
         parser.KNOWN_MULTI_FIELDS.append(CustomMultiplexingField)
         args = {"foos": CustomMultiplexingField()}
         result = parser.parse(args, web_request, location="query")
-        assert result["foos"] in ("a", "b")
+        assert result["foos"] in (["a", "b"], ["b", "a"])
```
lafrech, line 1120: I'm wondering why it passed before the change, how this change is related to the `Tuple` change.

lafrech (approved):
LGTM.

Just a question/comment but you can merge.

tests/test_core.py
sirosen: Ah, sorry, I forgot to mention this in the PR. It's not related to the Tuple change but is a fix for this test (which was broken prior to this PR).

I switched a line in the test parser:
```diff
-        return MultiDictProxy(req.query, schema)
+        return self._makeproxy(req.query, schema)
```

This test was actually incorrect before (!), but because I copied the wrong `assert` from one of the other cases and didn't fix the test parser, I didn't notice it when I added it.

Since `CustomMultiplexingField` is being added to the parser as a multi-field, it should return a list in this case.

PR #595, opened 2021-03-12
Modify unknown default to `None` for json, form and json_or_form

### LINKED ISSUE #580: [RFC] Modify unknown default to `None` for  json, form and json_or_form
Currently, we default to `EXCLUDE` for locations like query args, headers, cookies, files, where it is expected to get unknown args (or the user may parse args with multiple schemas and nested calls to use_args so `EXCLUDE` wouldn't be practical) and we default to RAISE for structures such as json or form request bodies where the input is unique and shouldn't contain unknown stuff.

The user may change the value to `EXCLUDE`, or even `None` to use the value from the schema.

However, because

- json and form schemas often include nested schemas
- unknown set by webargs parser doesn't propagate down to nested schemas, so it only applies to first level

the user wanting to set `EXCLUDE` must set it in the schema and in each nested schema (typically using a common base schema). But for this to work, he also needs to change the value in the webargs parser otherwise this value would take precedence other the schema value.

This makes me think that `None` would be a better default for these locations.

Since the schemas use `RAISE` by default, it wouldn't be an issue for the `RAISE` case. No more work for the user. But it would make it easier to use another value.

This whole "set unknown" feature was introduced because marshmallow's default of `RAISE` wasn't adapted to query args, headers, cookies and files. It was a bit painful to modify, especially when using `from_dict` and it deserved a better default, and that default didn't need to propagate (because those locations generally don't support nesting). But for nested structures, there's not much webargs can do and we should use marshmallow schema configuration (Meta attributes) to set the behaviour so having `None` as a default makes total sense.

The proposal is to change the default to

```py
    DEFAULT_UNKNOWN_BY_LOCATION: typing.Dict[str, str] = {
        "json": None,  # RAISE -> None
        "form": ma.None,  # RAISE -> None
        "json_or_form": ma.None,  # RAISE -> None
        "querystring": ma.EXCLUDE,
        "query": ma.EXCLUDE,
        "headers": ma.EXCLUDE,
        "cookies": ma.EXCLUDE,
        "files": ma.EXCLUDE,
}
```

This change is a breaking change if someone uses a schema with an unknown value set to a value other than `RAISE` on a json/form location but did not change the value in webargs. It doesn't seem to make much sense to intentionnaly set the schema unknown value to something other than `RAISE` while letting webargs set this back to `RAISE`, even if only for the top level schema.

Still, it may be considered a breaking change, so to be on the safe side, we can postpone this to the next major version.

lafrech:
Closes #580.

No test was broken and I didn't add any.

CHANGELOG will need to be sorted out before releasing but I did it this way to minimize merge conflicts.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,10 +1,13 @@
 Changelog
 ---------
 
-7.1.0 (Unreleased)
+8.0.0 (Unreleased)
 ******************
 
 Features:
+
+* ``unknown`` defaults to `None` for body locations (`json`, `form` and
+  `json_or_form`) (:issue:`580`).
 
 * Detection of fields as "multi-value" for unpacking lists from multi-dict
   types is now extensible with the `is_multiple` attribute. If a field sets
--- a/src/webargs/aiohttpparser.py
+++ b/src/webargs/aiohttpparser.py
@@ -71,7 +71,7 @@ del _find_exceptions
 class AIOHTTPParser(AsyncParser):
     """aiohttp request argument parser."""
 
-    DEFAULT_UNKNOWN_BY_LOCATION = {
+    DEFAULT_UNKNOWN_BY_LOCATION: typing.Dict[str, typing.Optional[str]] = {
         "match_info": RAISE,
         "path": RAISE,
         **core.Parser.DEFAULT_UNKNOWN_BY_LOCATION,
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -126,10 +126,10 @@ class Parser:
     DEFAULT_LOCATION: str = "json"
     #: Default value to use for 'unknown' on schema load
     #  on a per-location basis
-    DEFAULT_UNKNOWN_BY_LOCATION: typing.Dict[str, str] = {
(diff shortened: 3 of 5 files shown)


sirosen:
I'm not sure I quite understand the mypy failure.

It looks like it's failing in cases where we've subclassed and override `DEFAULT_UNKNOWN_BY_LOCATION` because it deduces the type as `Mapping[str, str]` and then we include `**Parser.DEFAULT_UNKNOWN_BY_LOCATION` which now is a subtype of `Mapping[str, Optional[str]]`...?

Maybe the solution is just to explicitly add a type annotation in the pyramid parser and the aiohttp parser.

The change itself looks good -- exactly what we talked about and you put forth as a better default -- we just need to get the linting to agree that it's good. 😉

lafrech:
Yeah, I don't get it either. I suspect a mypy issue but I'm not educated enough to even file a bug report.

> Maybe the solution is just to explicitly add a type annotation in the pyramid parser and the aiohttp parser.

It's late so I won't investigate any further now. I just did this and it passed.

sirosen (approved):
Looks good! 👍

PR #609, opened 2021-06-02
Add doc on upgrade to 8.0 from 7.0

### LINKED ISSUE #602: Add section to upgrading doc for v8.0
This was mentioned in an issue somewhere, but now I can't find it.

We should add a section to the doc, although it will probably be pretty short and simple in this case.

sirosen:
There are new features in webargs 8, but only one breaking change. (We could also consider treating `Tuple` as a "multiple field" breaking, but it's unlikely to impact most users.)

Therefore, the doc only needs to explain this one change: `unknown=None` by default for request body locations.

closes #602

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -3,6 +3,57 @@ Upgrading to Newer Releases
 
 This section documents migration paths to new releases.
 
+Upgrading to 8.0
+++++++++++++++++
+
+In 8.0, the default values for ``unknown`` were changed.
+When the location is set to ``json``, ``form``, or ``json_or_form``, the
+default for ``unknown`` is now ``None``. Previously, the default was ``RAISE``.
+
+Because ``RAISE`` is the default value for ``unknown`` on marshmallow schemas,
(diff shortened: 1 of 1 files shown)


docs/upgrading.rst
```diff
@@ -3,6 +3,57 @@ Upgrading to Newer Releases
...
+        ...
+
+
+In this case, under webargs 7.0 the schema ``unknown`` setting of ``EXCLUDE``
+would be ignored. Instead, ``unknown=RAISE`` would be used.
+
+In webargs 8.0, the schema ``unknown`` is used.
+
+To get the webargs 8.0 behavior (overriding the Schema ``unknown``), simply
+pass ``unknown`` to ``use_args``, as in
```
lafrech, line 49: I guess you meant 7.0 behavior

lafrech (approved):
Thanks for this!

Approved. Just a minor glitch to fix.

docs/upgrading.rst
sirosen: Yeah, good catch! I'll fix and then merge.

Issue #623, opened 2021-07-20
DelimitedList with empty string return ['']


TTWShell:
https://github.com/marshmallow-code/webargs/blob/ed8840c4bb06f1b3bcaf664e025183423349d9d2/src/webargs/fields.py#L72

```
In [8]: from webargs.fields import DelimitedList, Str

In [9]: dlist = DelimitedList(Str(), required=True)

In [10]: dlist.deserialize('')
Out[10]: ['']
```

Excepted empty list `[]` ?

lafrech:
I believe you're right. Would you like to propose a fix?

TTWShell:
> I believe you're right. Would you like to propose a fix?

I will do it

sirosen:
This was fixed in #624 and released. (Thanks @lafrech for doing that!)

PR #624, opened 2021-07-21
fix DelimitedList return [''] when field value is empty string

TTWShell:


TTWShell:
#623

sirosen:
Sorry for the slow turnaround on this -- I've been away from webargs for a little. Just caught up, and thanks for contributing the fix! 😄 
The fix looks good and I can't think of any weird corner cases. I'm going merge it.

@lafrech, should I do a point release with this, 8.0.1?

lafrech:
Yes, sorry about this. Saw it then forgot. Holidays. Thanks for the fix.

> @lafrech, should I do a point release with this, 8.0.1?

Absolutely.

Care to add @TTWShell to authors?

Thanks.

lafrech:
NVM, I did the release.

PR #647, opened 2021-10-18
Support Python 3.10

lafrech:
Also only test on min/max Python versions.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,9 +29,6 @@ jobs:
       - mypy
       - py36
       - py36-mindeps
-      - py37
-      - py38
-      - py39
       - py310
       - py310-marshmallowdev
       - docs
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,7 +32,8 @@ jobs:
       - py37
       - py38
       - py39
-      - py39-marshmallowdev
+      - py310
+      - py310-marshmallowdev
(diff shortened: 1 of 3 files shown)


sirosen (approved):
I'm onboard with reducing the CI matrix.

Should we also add a 3.11-dev build?

lafrech:
> Should we also add a 3.11-dev build?

No strong objection but I'm not sure it's worth the CPU.

A new Python version gets published once every year. We (at least I) generally look at the changes before it gets out out of curiosity and there is almost no breaking change, and we generally add it to the CI rapidly after it is out. It's not like we expect any surprise we'd really need to anticipate.

We could add it once it reaches beta. In fact, I don't know if CIs (azure, Github) provide it before beta anyway.

sirosen:
That's all very fair. Let's wait on pythons being released to test, as we did with 3.10 .
GitHub supports a huge list of pythons, including 3.11 today. I don't know about Azure, but I assume it's similar.

I'm going ahead and merging.

lafrech:
Good.

BTW, I find Azure a hell to configure and I switched to Github for flask-smorest. Perhaps we could do the same for other marshmallow libs.

sirosen:
I've been pretty happy with GitHub Actions myself, and use it a lot at work. If only for my own level of familiarity with it, I'd be happy with a switch to it.

PR #650, opened 2021-11-08
Add py.typed to MANIFEST.in

lafrech:
Similar to https://github.com/marshmallow-code/marshmallow/pull/1905.


sirosen:
Yep, the issue trail in marshmallow makes sense.

PR #661, opened 2021-12-04
Add Flask async view support.

Isira-Seneviratne:
Add support for async Flask views, which were introduced in Flask 2.0.

### DIFF
--- a/src/webargs/flaskparser.py
+++ b/src/webargs/flaskparser.py
@@ -23,11 +23,10 @@ Example: ::
 import typing
 
 import flask
+import marshmallow as ma
 from werkzeug.exceptions import HTTPException
 
-import marshmallow as ma
-
-from webargs import core
+from webargs import core, asyncparser
 
 
 def abort(http_status_code, exc=None, **kwargs):
@@ -117,6 +116,18 @@ class FlaskParser(core.Parser):
         return flask.request
 
 
+class FlaskAsyncParser(FlaskParser, asyncparser.AsyncParser):
+    """Flask request argument parser for async views (supported on Flask
+    2.0 and higher).
+    """
+
+    pass
+
+
 parser = FlaskParser()
+async_parser = FlaskAsyncParser()
+
 use_args = parser.use_args
+use_args_async = async_parser.use_args
 use_kwargs = parser.use_kwargs
+use_kwargs_async = async_parser.use_kwargs
--- a/src/webargs/flaskparser.py
+++ b/src/webargs/flaskparser.py
@@ -121,8 +121,6 @@ class FlaskAsyncParser(FlaskParser, asyncparser.AsyncParser):
     2.0 and higher).
     """
 
-    pass
-
 
 parser = FlaskParser()
 async_parser = FlaskAsyncParser()
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -55,3 +55,4 @@ Contributors (chronological)
 * `@michaelizergit <https://github.com/michaelizergit>`_
 * Legolas Bloom `@TTWShell <https://github.com/TTWShell>`_
 * Kevin Kirsche  `@kkirsche <https://github.com/kkirsche>`_
+* Isira Seneviratne `@Isira-Seneviratne <https://github.com/Isira-Seneviratne>`_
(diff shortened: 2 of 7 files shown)


lafrech:
@Isira-Seneviratne did you close this PR by accident?

I didn't review it yet since it is closed.

We're about to drop Python 3.6 so CI errors should disappear.

Isira-Seneviratne:
> @Isira-Seneviratne did you close this PR by accident?
> 

No, I closed because of the errors in CI.

> I didn't review it yet since it is closed.
> 
> We're about to drop Python 3.6 so CI errors should disappear.

Okay, great. I'll reopen this PR.

src/webargs/flaskparser.py
```diff
@@ -117,6 +116,18 @@ def get_default_request(self):
         return flask.request
 
 
+class FlaskAsyncParser(FlaskParser, asyncparser.AsyncParser):
+    """Flask request argument parser for async views (supported on Flask
+    2.0 and higher).
+    """
+
+    pass
```
sirosen:
```suggestion
```

This is a cosmetic nitpick, but I'd prefer we leave the body empty rather than an unnecessary `pass`.

src/webargs/flaskparser.py
```diff
@@ -117,6 +116,18 @@ def get_default_request(self):
         return flask.request
 
 
+class FlaskAsyncParser(FlaskParser, asyncparser.AsyncParser):
```
sirosen, line 119: I'd like it if we could, in the future, support async flask views in `FlaskParser` itself, but I foresee issues with it (primarily that `FlaskParser.parse` needs to be synchronous). This approach of adding a new parser class is easier to do now and gets us some support. We can always revisit in v9 or later.
👍 for thinking of this approach!

I need to think about this more, but I'm pretty sure this empty class is correct.
The MRO should be `[FlaskParser, AsyncParser, Parser, object]` from this, and that seems like it will behave correctly.

sirosen (commented):
First of all, a huge thanks for the contribution! I think the approach of a new parser to handle the flask async case is the right one.

The new tests look good, and I'm going to start reviewing them more carefully to make sure there are no odd cases we need to consider which aren't covered.

I want to get this passing tests, which means fixing the tests on `dev`. I'm working on fixing that right now.

The only other thing that I think we need is some kind of update to the documentation to note that using `webargs` on a flask async view requires `use_args_async`, `use_kwargs_async`, `async_parser`, or `FlaskAsyncParser`. Could you add some lines to docs/framework_support.rst` to cover this? It doesn't need to be comprehensive -- just something similar to the flask example which is already there, but for an async route.

src/webargs/flaskparser.py
Isira-Seneviratne: > The only other thing that I think we need is some kind of update to the documentation to note that using `webargs` on a flask async view requires `use_args_async`, `use_kwargs_async`, `async_parser`, or `FlaskAsyncParser`. Could you add some lines to docs/framework_support.rst` to cover this? It doesn't need to be comprehensive -- just something similar to the flask example which is already there, but for an async route.

Yeah, I can do that.

Isira-Seneviratne: > This is a cosmetic nitpick, but I'd prefer we leave the body empty rather than an unnecessary `pass`.

The `pass` statement is needed for empty classes, otherwise an error will occur.

sirosen: On what python versions? I find that this works fine on every interpreter I test:

```
$ cat foo.py
class A:
    """hi"""
$ for py in python3.6 python3.7 python3.8 python3.9 python3.10 python3.11 pypy3; do $py foo.py; done
# no output, no errors
```

The class has a body if it has a docstring, which is sufficient.

Isira-Seneviratne: Oh, right. Thanks!

src/webargs/flaskparser.py
lafrech: @slallema said in #662 they did it by modifying FlaskParser. Details about how would be welcome.

lafrech:
@Isira-Seneviratne @slallema would you like to pick this up now that Python 3.6 is dropped and CI is fixed?

No rush. It does not have to make it into 8.1.0. It can wait until a later release.

sirosen:
I'm closing this because this was included in / served as the basis for #701 .

@Isira-Seneviratne was credited in the changelog for the initial work, and we'll find an appropriate time to release `webargs` v8.2.0 with this. I hope everyone ends up happy with this!

lafrech:
Superseded by #701.

PR #664, opened 2021-12-12
Fix Django usage in testsuite

sirosen:
To be compatible with the latest versions of django, `django.conf.urls.url` is no longer supported. Use `django.urls.re_path` as a replacement.

This gets the tests passing for me locally again. I'm hoping it works cleanly in Azure as well.
I also see a bunch of deprecation warnings I'd like to address after this.


lafrech (approved):
LGTM.

Tests pass and "mindeps" target in tox ensures this already existed in Django 2.2.0.

PR #673, opened 2022-01-02
Drop Python 3.6

lafrech:
https://endoflife.date/python


lafrech:
I think this can be added to 8.1.0.

azure-pipelines.yml
```diff
@@ -27,8 +27,7 @@ jobs:
     toxenvs:
       - lint
       - mypy
-      - py36
```
sirosen, line 30: Should we keep `py37` in the list? Right now, it's `py36,py36-mindeps`.

It's unlikely that `py37` would fail while `py37-mindeps` and `py310` are both passing, but I can imagine it happening in weird scenarios.

src/webargs/tornadoparser.py
```diff
@@ -57,8 +57,8 @@ def __getitem__(self, key):
                 return _unicode(value)
             return value
         # based on tornado.web.RequestHandler.decode_argument
-        except UnicodeDecodeError:
-            raise HTTPError(400, f"Invalid unicode in {key}: {value[:40]!r}")
+        except UnicodeDecodeError as exc:
+            raise HTTPError(400, f"Invalid unicode in {key}: {value[:40]!r}") from exc
```
sirosen, line 61: I'm just curious, was pyupgrade the tool that flagged `raise` vs `raise ... from`? I see a few of these, so I assume it was part of the linting update.

sirosen (commented):
I'm onboard if you want to merge, but I have a question about the CI config. I'd like us to agree on that.

azure-pipelines.yml
lafrech: It should. My mistake. Let me fix this.

src/webargs/tornadoparser.py
lafrech: IIRC, flake8 complained.

sirosen: Ah, I didn't notice the change in `flake8-bugbear`. I see now they introduced a new error, `B904` for this. Cool!

sirosen (approved):
Thanks for clearing that up! Let's merge.

Issue #667, opened 2021-12-16
typing issue with __version_info__ += __parsed_version__.pre


lafrech:
mypy issue:

```
    __version_info__ += __parsed_version__.pre
```

```
src/webargs/__init__.py:14: error: Unsupported operand types for + ("Tuple[int, ...]" and "Tuple[str, int]")
```

Not sure what the problem is. I'm tempted to just add a `# type: ignore`. Any better idea, anyone?

sirosen:
I think this might be a real issue?

Looking at the annotations in `packaging`, I see that `release` is `(int, ...)` and `pre` is `(str, int)`, as the error says.

I'm going to poke at the history in `packaging` (and here + in `marshmallow`'s `__init__.py`, where I presume something similar is done) to try to figure this out.

sirosen:
Okay, I think this and the originating PR in `marshmallow` both had a bug.
I'm going to put in a PR in `marshmallow` which solves this there, and (assuming we are satisfied with it), we'll take the same fix here. I need to figure out what we expect to get in `__version_info__` in this case.

sirosen:
As detailed in the marshmallow PR, I had mixed up the semantics of `+` on tuples. (I thought it was vector addition, not concatenation.) So the code is correct, but we need to either annotate the version_info tuple as `tuple[str | int, ...]` or as `tuple[int, int, int] | tuple[int, int, int, str, int]`. I'll continue to use the marshmallow PR to drive this, and we can just follow the decision we reach there.

lafrech:
Yeah, I didn't see any issue with tuple addition, hence the temptation to just silent the warning.

Of course, a more specific typing is better if you manage to achieve it.

lafrech:
~Perhaps we could pin mypy to 0.910 in the meantime to avoid CI errors on unrelated PRs.~

Edit: In fact it is still pinned to 0.910 in dev branch as we haven't merged #670 yet. I had it bumped to 0.930 through `pre-commit autoupdate` while working on #673. I just downgraded it in that branch to let tests pass.

Edit 2: It is pinned in pre-commit config but not in tox in the dedicated mypy target. I sent #674 to pin it there too.

sirosen:
@lafrech, thanks for taking care of that!

Do you have any opinion about https://github.com/marshmallow-code/marshmallow/pull/1925 ?
I can change it to `__version_info__: tuple[str | int, ...] = ...` (which is the smallest fix) but I still think the change I proposed is slightly nicer.

PR #674, opened 2022-01-02
Pin mypy version in tox

lafrech:
Currently pinning to 0.910 due to #667.


lafrech:
Let's merge this for now until we find the proper fix for #667.

PR #680, opened 2022-01-04
Fix type annotations for `__version_info__` + update mypy

### LINKED ISSUE #667: typing issue with __version_info__ += __parsed_version__.pre

sirosen:
Mirrors the work in https://github.com/marshmallow-code/marshmallow/pull/1925

Resolves #667, #670

With this fixed, the latest mypy is usable again, so I've included the update.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     additional_dependencies: [black==21.12b0]
     args: ["--target-version", "py37"]
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.910-1
+  rev: v0.930
   hooks:
   - id: mypy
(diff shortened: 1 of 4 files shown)


lafrech (approved):
I think the mypy issue was an error in mypy that was fixed in mypy 0.931 (#683), although I couldn't find the changelog entry explaining what was fixed.

Still, this change makes the typing more specific without being too heavy (compared to the earlier PR in marshmallow involving a helper method).

PR #681, opened 2022-01-04
Rename `Nested` to `WebargsNested`

### LINKED ISSUE #677: Should `webargs.fields.Nested` be renamed/aliased to avoid confusion with `marshmallow.fields.Nested`?
## Context

`webargs.fields` provides all of the names from `marshmallow.fields`, but it rewrites `Nested`:
https://github.com/marshmallow-code/webargs/blob/0b23a8dd3bf464ca447d759fdbdc373f97666fb7/src/webargs/fields.py#L26

This is the only field which changes behavior between the two `fields` modules, since there are no other redefinitions. It's possible that `marshmallow` introduces something named `DelimitedList` or other conflicting names in the future, but (1) that's not very likely to happen in practice (2) we could deal with it when it happens. So, focusing on `Nested`...

#671 is an instance of type confusion surrounding this. I'm not yet sure if it's user error or if we've confused a type checker.

## Proposed Resolution

Rename `Nested` (gradually) in `webargs.fields`. Possible names: `WebargsNested`, `WNested`, `NestedArg`, `DictNested` (because it accepts a dict).

This can be done by renaming the class and assigning, e.g. `Nested = WebargsNested`.
A deprecation warning can be added later for the old name when passed a dict, prior to a major version which drops it.

## (rejected) Alternative Solutions

I'm happy to be convinced that one of these is better than my proposal. But these are what I considered.

### Do nothing

Always an option. But I look at that ignored redefinition type error as a warning sign that we could trip over this in the future.
Plus, it is confusing that `Nested` doesn't quite match the `marshmallow` docs.

### Give `marshmallow.fields.Nested` the ability to accept a dict

Obsoletes `webargs.fields.Nested`.

This is very simple, and if dict-based definition is useful to `webargs` users, maybe it's useful to all `marshmallow` users.
However, I'd only pursue this with buy-in from the `marshmallow` maintainers.

### Drop support for dict-based definition in `webargs.fields.Nested`

Basically, remove the custom subclass. It's obviously the "best" for maintenance.

However, it's probable that users are using and like this feature. When we considered removing all built-in support for dict definition in the past, users informed us that they find it useful. The utility of dict definition _might_ not hold for `Nested` in the same way as simple query param parsing, but I have no evidence for that.

sirosen:
This renames the existing `Nested` subclass in `webargs.fields` to `WebargsNested`. This allows type checkers to identify it as a
different type from `Nested` (and therefore not throw errors when a dict is passed), and helps clarify that we have slightly modified `Nested`.

In order to remain compatible with existing code which uses `Nested` to handle dicts, a new `Nested` class is defined which throws a DeprecationWarning when used on a `dict`.

The end result is that any usage of `Nested` which would work with `marshmallow.fields.Nested` will not warn. Conversely, any usage which is not allowed (the dict usage) will warn. We should be able to remove the `Nested` class from `webargs.fields` in a future major release.

All internal docs and test apps are updated to use `WebargsNested`.
The new `Nested` shim is not publicly documented, as it should only be needed for compatibility. There is a small note added to the `WebargsNested` docstring stating that it used to be named `Nested`. This should help alleviate confusion when users are reading code which still uses `webargs.fields.Nested({...})` and referencing the latest docs.

resolves #677

---

I'm still (more than) happy to go with a different name, or to ditch this PR and take a different approach. It seems that `mypy` can't be convinced that `webargs.fields.Nested` takes a dict -- I tried type annotating `__init__` and it still showed errors on `Nested` usage -- so I want to do something.

Probably the trickiest part of this, IMO, was figuring out how to do it such that the docs don't become confusing. The note in `WebargsNested` and no direct mention of the `Nested` redefinition is the best I came up with.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -14,7 +14,7 @@ webargs.fields
 --------------
 
 .. automodule:: webargs.fields
-    :members: Nested, DelimitedList
+    :members: WebargsNested, DelimitedList, DelimitedTuple
 
 
 webargs.multidictproxy
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -228,7 +228,7 @@ Nesting Fields
     from webargs import fields
 
     args = {
-        "name": fields.Nested(
+        "name": fields.WebargsNested(
             {"first": fields.Str(required=True), "last": fields.Str(required=True)}
         )
     }
--- a/src/webargs/fields.py
+++ b/src/webargs/fields.py
@@ -15,6 +15,8 @@ tells webargs where to parse the request argument from.
 """
 from __future__ import annotations
 
+import warnings
+
 import marshmallow as ma
 
 # Expose all fields from marshmallow.fields.
@@ -23,9 +25,12 @@ from marshmallow.fields import *  # noqa: F40
 __all__ = ["DelimitedList", "DelimitedTuple"] + ma.fields.__all__
 
 
-class Nested(ma.fields.Nested):  # type: ignore[no-redef]
+class WebargsNested(ma.fields.Nested):
     """Same as `marshmallow.fields.Nested`, except can be passed a dictionary as
     the first argument, which will be converted to a `marshmallow.Schema`.
+
+    In previous versions, this field was named ``Nested`` and replaced ``Nested``
+    from ``marshmallow.fields``.
 
     .. note::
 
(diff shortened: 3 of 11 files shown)


sirosen:
With https://github.com/marshmallow-code/marshmallow/pull/1935 merged, in going to close this.

I think that once there's a marshmallow release we might want a change here, but I'm not certain yet. This approach is not what we want though.

PR #688, opened 2022-02-06
Remove 'lint' from azure-pipelines

sirosen:
With pre-commit.ci , it should not be necessary to run this job in Azure Pipelines.

---

This is a very simple change, meant in part to trigger the pre-commit.ci integration and make sure that it works.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,7 +25,6 @@ jobs:
 - template: job--python-tox.yml@sloria
   parameters:
     toxenvs:
-      - lint
       - mypy
       - py37
       - py37-mindeps


sloria (approved):
Nice. Can you also make the pre-commit.ci step a required check in the branch protection settings?

sirosen:
I marked `pre-commit.ci` as required. The UI didn't show me clearly if the Azure checks are required or not right now (there was nothing listed there), but I'll leave any change to that as a separate matter.

lafrech:
> I marked pre-commit.ci as required. The UI didn't show me clearly if the Azure checks are required or not right now (there was nothing listed there), but I'll leave any change to that as a separate matter.

Is this still the case? I don't see any check required in branch protection.

(I'm trying to replicate the config here in marshmallow repo.)

sirosen:
I don't see it enabled there when I look. Perhaps I did it incorrectly at the time?

We should mark it required as requested/discussed before, which I'll go and do now. In a few minutes, it would be good for someone else to look at the config (as I leave it) to check my work.

lafrech:
I don't see anything in marshmallow settings either. Yet, when CI runs, the UI allows me to merge if all tests pass and warns in red if I "use admin privileges" to merge before they pass or after they fail.

Thinking of it, this behaviour suits me.

Issue #573, opened 2020-12-21
webargs 6.x -> 7.x - NoReturn issue in python 3.7.1


AbdealiLoKo:
When I upgraded my webargs from `6.1.1` to `7.0.1`, I started getting this error:

```python
11:54:14 venv/lib/python3.7/site-packages/webargs/__init__.py:7: in <module>
11:54:14     from webargs.core import ValidationError
11:54:14 venv/lib/python3.7/site-packages/webargs/core.py:33: in <module>
11:54:14     ErrorHandler = typing.Callable[..., typing.NoReturn]
11:54:14 /opt/miniconda3/lib/python3.7/typing.py:755: in __getitem__
11:54:14     return self.__getitem_inner__(params)
11:54:14 /opt/miniconda3/lib/python3.7/typing.py:251: in inner
11:54:14     return func(*args, **kwds)
11:54:14 /opt/miniconda3/lib/python3.7/typing.py:774: in __getitem_inner__
11:54:14     result = _type_check(result, msg)
11:54:14 /opt/miniconda3/lib/python3.7/typing.py:135: in _type_check
11:54:14     raise TypeError(f"Plain {arg} is not valid as type argument")
11:54:14 E   TypeError: Plain typing.NoReturn is not valid as type argument
```

On searching around I found that this issue is most likely due to: https://bugs.python.org/issue34921
And I was using python 3.7.0 - which is affected.

Just thought I'd bring it up here, that webargs 7.x won't work with python 3.7.0 and 3.7.1

lafrech:
I don't think there's anything we should do about this in webargs since the bug was in Python and was fixed, but thanks for sharing this, it may help someone facing the same issue.

kasium:
@lafrech I just also hit this issue and switching to another python version is kind of hard (various reasons).
Therefore I would like to ask if you would be open to accept a simple PR which just adds a simple `if` to avoid this issue.

If you don't think that this makes sense, you might want to consider to update the `python_requires`

sirosen:
> If you don't think that this makes sense, you might want to consider to update the `python_requires`

I think we'd welcome a patch to our `python_requires` to `>=3.7.2`.
This actually happened to us (also `NoReturn`, IIRC) in python 3.6.0 and we set `python_requires>=3.6.2` to fix it.

IMO `python_requires` is the only appropriate way to handle this. Conditional logic for type annotations can have some nasty implications.

kasium:
@sirosen thanks, I'll create a PR

PR #692, opened 2022-02-22
Set python_requires to 3.7.2

kasium:
See https://github.com/marshmallow-code/webargs/issues/573 for reasons

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,7 @@ setup(
         "api",
         "marshmallow",
     ),
-    python_requires=">=3.7",
+    python_requires=">=3.7.2",
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Intended Audience :: Developers",


sirosen:
Thanks for submitting this patch!
I think this is good, but I'd like to wait in case @lafrech sees anything wrong with doing this before we merge.

---

The build failure is due to a change in the `pallets` libraries which breaks clean installs of some of their EOL versions. We're testing against an old `flask` version, which is why it fails. I'll think about how to fix this when I'm able, hopefully later today.

kasium:
> Thanks for submitting this patch! I think this is good, but I'd like to wait in case @lafrech sees anything wrong with doing this before we merge.
> 
> The build failure is due to a change in the `pallets` libraries which breaks clean installs of some of their EOL versions. We're testing against an old `flask` version, which is why it fails. I'll think about how to fix this when I'm able, hopefully later today.

Thanks!

lafrech (approved):
Good. Thanks.

sirosen:
I just rebased this onto `dev` now that we've fixed the build issue. I expect it to build cleanly and then we can merge.

Once this is in, I'll want to pull together v8.1.1 with this fix and attribution for this PR in the changelog. Maybe it will be our first release published via GitHub Actions.

PR #694, opened 2022-02-27
Pin markupsafe and itsdangerous in mindeps build

sirosen:
Because `flask` does not pin its dependencies, it is necessary to include at least some pins when trying to install any version other than the latest. A comment in the tox config explains this briefly.

In order to also creep forward the lowest supported version (0.12.5 is *very* old), use one of the earlier flask 1.x releases, along with versions of markupsafe and itsdangerous which are approximately contemporaneous with it.

Whether or not these versions are the best choice, this set works and should get the build passing again.

---

We don't publicly announce what versions of the various frameworks we support, but this change at least makes sure we're testing on "flask 1.x" .

As a bit of separate follow-up, I've long been somewhat "low key nervous" about the fact that we install all of the frameworks in a single env for our tests. I want to look at turning that into a matrix in tox, with each framework as a factor, to see if it works out nicely. I might look at that after the GitHub Actions change.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,12 @@ extras = tests
 deps =
     !marshmallowdev: marshmallow>=3.0.0,<4.0.0
     marshmallowdev: https://github.com/marshmallow-code/marshmallow/archive/dev.tar.gz
-    mindeps: Flask==0.12.5
+# for 'mindeps', list flask dependencies to ensure a successful install
+# because flask 1.x does not set upper bounds on the versions of its pallets dependencies
+    mindeps: Flask==1.0.4
+    mindeps: markupsafe==1.1.1
+    mindeps: itsdangerous==1.1.0
+# all non-flask frameworks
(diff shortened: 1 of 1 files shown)


lafrech (commented):
One might argue that this means dropping Flask < 1.x support but let's not bother. Most libs don't even test against old dependencies versions.

We could use Flask 1.1.3 to avoid listing the dependencies.

https://github.com/pallets/flask/pull/4047

sirosen:
I updated to `flask==1.1.3`, but the `markupsafe` pin needs to remain in place. I've expanded the comment to explain the situation, with links to this PR, and a couple of relevant pallets issues.

> One might argue that this means dropping Flask < 1.x support but let's not bother. Most libs don't even test against old dependencies versions.

Yeah, I don't think we need to announce anything. We're unusual in that we're even trying to do this test.

lafrech (approved):

PR #695, opened 2022-03-02
Remove testpypi publish support

sirosen:
I just tried this out with the `testpypi-8.1.1` tag, and it worked!
image

If you open up that build output, `release` pushed to testpypi as intended.

Now that the pipeline has been tested, there's no need for a variant which publishes to testpypi. I'd be fine with keeping it around, but it may just end up being something to trip over in the future.

---

I'm going to delete the `testpypi-8.1.1` tag when/if this merges, unless anyone has a reason that I should not do so.
(I mentioned this elsewhere, but this PR seems like a better place to discuss.)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -58,11 +58,5 @@ jobs:
         run: python -m build
       - name: check package metadata
         run: twine check dist/*
-      # tags starting with 'testpypi-' publish to test pypi
-      # those without that prefix publish to production pypi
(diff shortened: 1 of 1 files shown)


lafrech (approved):
Good. Thanks.

> I'm going to delete the testpypi-8.1.1 tag when/if this merges, unless anyone has a reason that I should not do so.

Sure.

sloria:
awesome!

PR #701, opened 2022-03-16
Merge AsyncParser into Parser, completing Flask async support

sirosen:
I only just put this together today, so I want to share it, but also think a little more about whether or not this is a good idea. I want to read over the various framework parsers (maybe tomorrow) to see if there are any surprises which the tests don't catch. This also could impact `webargs-starlette`, so I might take a look there too.

1. Rebase #661 

2. Apply a change on top of it, https://github.com/marshmallow-code/webargs/commit/a4fe5546950e693182eae19002743263372fc02f

The idea is to combine `Parser` and `AsyncParser` into a single class (`Parser`). I think some of the code duplication we have today comes from the fact that async support isn't built into the base class.

Almost everything works out smoothly. The only odd case was that I saw some cases where the `mock.patch` decorator didn't play nicely with async tests, so those tests use the `mock.patch` context manager instead.

### DIFF
--- a/docs/framework_support.rst
+++ b/docs/framework_support.rst
@@ -13,8 +13,8 @@ Flask support is available via the :mod:`webargs.flaskparser` module.
 Decorator Usage
 +++++++++++++++
 
-When using the :meth:`use_args <webargs.flaskparser.FlaskParser.use_args>` and :meth:`use_args <webargs.flaskparser.FlaskParser.use_args_async>`
-decorators, the arguments dictionary will be *before* any URL variable parameters.
+When using the :meth:`use_args <webargs.flaskparser.FlaskParser.use_args>`
+decorator, the arguments dictionary will be *before* any URL variable parameters.
 
 .. code-block:: python
 
@@ -25,13 +25,6 @@ decorators, the arguments dictionary will be *before* any URL variable parameter
     @app.route("/user/<int:uid>")
     @use_args({"per_page": fields.Int()}, location="query")
     def user_detail(args, uid):
-        return ("The user page for user {uid}, showing {per_page} posts.").format(
-            uid=uid, per_page=args["per_page"]
-        )
-
-    @app.route("/user_async/<int:uid>")
-    @use_args_async({"per_page": fields.Int()}, location="query")
-    async def user_detail_async(args, uid):
         return ("The user page for user {uid}, showing {per_page} posts.").format(
             uid=uid, per_page=args["per_page"]
         )
--- a/docs/framework_support.rst
+++ b/docs/framework_support.rst
@@ -13,7 +13,8 @@ Flask support is available via the :mod:`webargs.flaskparser` module.
 Decorator Usage
 +++++++++++++++
 
-When using the :meth:`use_args <webargs.flaskparser.FlaskParser.use_args>` decorator, the arguments dictionary will be *before* any URL variable parameters.
+When using the :meth:`use_args <webargs.flaskparser.FlaskParser.use_args>` and :meth:`use_args <webargs.flaskparser.FlaskParser.use_args_async>`
+decorators, the arguments dictionary will be *before* any URL variable parameters.
 
 .. code-block:: python
 
@@ -24,6 +25,13 @@ When using the :meth:`use_args <webargs.flaskparser.FlaskParser.use_args>` decor
     @app.route("/user/<int:uid>")
     @use_args({"per_page": fields.Int()}, location="query")
     def user_detail(args, uid):
+        return ("The user page for user {uid}, showing {per_page} posts.").format(
+            uid=uid, per_page=args["per_page"]
+        )
+
+    @app.route("/user_async/<int:uid>")
+    @use_args_async({"per_page": fields.Int()}, location="query")
+    async def user_detail_async(args, uid):
         return ("The user page for user {uid}, showing {per_page} posts.").format(
             uid=uid, per_page=args["per_page"]
         )
--- a/examples/flask_example.py
+++ b/examples/flask_example.py
@@ -17,7 +17,7 @@ import datetime as dt
 
 from flask import Flask, jsonify
 from webargs import fields, validate
-from webargs.flaskparser import use_args, use_kwargs, use_kwargs_async
+from webargs.flaskparser import use_args, use_kwargs
 
 app = Flask(__name__)
 
@@ -42,9 +42,9 @@ def add(x, y):
 
 
 @app.route("/subtract", methods=["POST"])
-@use_kwargs_async(add_args)
+@use_kwargs(add_args)
 async def subtract(x, y):
-    """An async addition endpoint."""
+    """An async subtraction endpoint."""
     return jsonify({"result": x - y})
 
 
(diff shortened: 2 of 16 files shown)


sirosen:
I think this change is good after reviewing various framework docs, and I'm increasingly confident that it's the right approach.
The new parser behavior handles async views in django and falcon, and tests are passing for me.

I realized after pushing that my changes to add falcon and django async tests broke the `py37-mindeps` build. I'll work on cleaning that up later, but it should be possible to make the tests skip in that case like I did with the flask async tests.

docs/framework_support.rst
```diff
@@ -28,6 +29,14 @@ When using the :meth:`use_args <webargs.flaskparser.FlaskParser.use_args>` decor
             uid=uid, per_page=args["per_page"]
         )
 
+
+    @app.route("/user_async/<int:uid>")
+    @use_args_async({"per_page": fields.Int()}, location="query")
```
sirosen: I need to remember to come back and fix this example (and the line above).

With the way that I changed things, `use_args` handles sync and async.

examples/flask_example.py
```diff
@@ -40,6 +41,13 @@ def add(x, y):
     return jsonify({"result": x + y})
 
 
+@app.route("/subtract", methods=["POST"])
+@use_kwargs_async(add_args)
```
sirosen: Like `use_args_async` -> `use_args`, this can be `use_kwargs`.

src/webargs/asyncparser.py
```diff
@@ -1,25 +1,17 @@
...
 
 class AsyncParser(core.Parser):
-    """Asynchronous variant of `webargs.core.Parser`, where parsing methods may be
-    either coroutines or regular methods.
+    """Asynchronous variant of `webargs.core.Parser`.
+
+    The ``parse`` method is redefined to be ``async``.
     """
 
-    # TODO: Lots of duplication from core.Parser here. Rethink.
```
sirosen, line 22: ❗

Rethought!

I've been staring at AsyncParser pretty much since the first commits time I looked at webargs source. I think this might finally be a solution we can live pretty happily with? And in v9.0 or later we could actually remove `AsyncParser`.

src/webargs/core.py
```diff
@@ -267,6 +308,61 @@ def _get_schema(self, argmap: ArgMap, req: Request) -> ma.Schema:
...
+                if self.unknown != _UNKNOWN_DEFAULT_PARAM
+                else self.DEFAULT_UNKNOWN_BY_LOCATION.get(location)
+            )
+        )
+        load_kwargs: dict[str, typing.Any] = {"unknown": unknown} if unknown else {}
+        preprocessed_data = self.pre_load(
+            location_data, schema=schema, req=req, location=location
+        )
+        data = schema.load(preprocessed_data, **load_kwargs)
+        self._validate_arguments(data, validators)
```
sirosen, line 363: Right now, several of these methods are still synchronous-only. But if we take this model further, we could allow
```python
class MyParser(DjangoParser):
    async def pre_load(...): ...
```
and properly detect and `await` that in here.

It would bulk up the amount of code we'd have to duplicate, but with some nice structure in place for it, we might not find it too difficult. It's not essential and not supported today, even by the AsyncParser, so I don't see it as necessary.

sirosen (commented):
I just took another look at this and noticed some doc examples which need to be tweaked.

src/webargs/flaskparser.py
```diff
@@ -23,9 +23,8 @@ def user_detail(args, uid):
 from __future__ import annotations
 
 import flask
-from werkzeug.exceptions import HTTPException
-
 import marshmallow as ma
+from werkzeug.exceptions import HTTPException
```
lafrech: This could have stayed just under `import flask`.

src/webargs/core.py
```diff
@@ -198,19 +201,23 @@ def _get_loader(self, location: str) -> typing.Callable:
...
         Needs to have the schema in hand in order to correctly handle loading
         lists from multidict objects and `many=True` schemas.
         """
         loader_func = self._get_loader(location)
-        data = loader_func(req, schema)
-        # when the desired location is empty (no data), provide an empty
-        # dict as the default so that optional arguments in a location
-        # (e.g. optional JSON body) work smoothly
-        if data is missing:
-            data = {}
```
lafrech, line 213: What's the rationale for removing this?

src/webargs/flaskparser.py
```diff
@@ -118,5 +117,6 @@ def get_default_request(self):
 
 
 parser = FlaskParser()
+
```
lafrech: Empty line added through subsequent edits while working on this PR. Remove it and move the werkzeug import up and this file is actually untouched by the whole PR.

lafrech (approved):
I finally took a bit of time to review this. Sorry about the delay.

This looks all good to me. The diff was scary but your comprehensive comments and explanations and a commit-per-commit review approach made the whole thing much easier.

My only comments are nitpicks you may want to consider if you're a edit-until-each-commit-is-perfect-enough-for-History maniac like me.

Commit message typo:
- try to consolidate and reduce duplicaiton between
+ try to consolidate and reduce duplication between

Also there's the use of `use_args_async` you want to fix in the docs (see your own comments).

Thanks for handling this and sorry again about the delay.

src/webargs/core.py
sirosen: It was moved into `_process_location_data`. Throughout the changes, I tried to keep the async vs sync copies of methods as short as possible. If they had a common piece of logic -- like this -- it moved to a new location if possible/feasible.

src/webargs/flaskparser.py
sirosen: I have to reread this PR a bit. I tried to revert any changes from the base PR which became irrelevant under my rewrite, but I definitely missed a couple.

src/webargs/core.py
lafrech: Oh, right. I should have `Ctrl+F`ed that

src/webargs/flaskparser.py
lafrech: No pb. It's mostly cosmetic. I tend to spend too much time struggling with `git rebase -i` to even have clean intermediate commits, for History (although nobody will ever study _that_ History).

sirosen:
> My only comments are nitpicks you may want to consider if you're a edit-until-each-commit-is-perfect-enough-for-History maniac like me.

I'm glad I'm not alone in doing this! 😁 

I've just done a rebase to tweak that commit message and fixup the whitespace changes. I think this is now good, but I'm going to do a final proofread.

sirosen:
I'll try to make time after work today for webargs. Should we do a release to get this into people's hands? I can take care of it if you're onboard, @lafrech.

lafrech:
The two last falcon tests are failing after pulling in my virtualenv but they disappear in a new virtualenv. Same falcon version so it must be in another lib.

    There is no current event loop in thread 'MainThread'.

No time right now to investigate. Probably an issue on my side.

sirosen:
Okay, why don't we wait a little before a release, in that case?

I wanted to add some more doc on this anyway, to demonstrate async support.

lafrech:
Good. No rush.

Meanwhile, interested users @Isira-Seneviratne and @slallema may give it a try pulling dev.

PR #721, opened 2022-05-02
Always return files location data for Flask parser

### LINKED ISSUE #720: When setting location to `files`, custom field's `_deserialize` and `_validated` methods were not called
Hi! I created a custom `File` field for file uploads:

```py
from werkzeug.datastructures import FileStorage
from marshmallow import Field


class File(Field):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.metadata['type'] = 'string'
        self.metadata['format'] = 'binary'

    default_error_messages = {
        'invalid': 'Not a valid file.'
    }

    def _deserialize(self, value, attr, data, **kwargs) -> t.Any:
        if not isinstance(value, FileStorage):
            raise self.make_error('invalid')
        return value
```

Then use it like this in a Flask application:

```py
from flask import Flask
from webargs import fields
from webargs.flaskparser import use_args

app = Flask(__name__)


class UploadShema(Schema):
    photo = File()

@app.route('/', methods=['post'])
@use_args(UploadSchema, location='files')
def index(args):
    ...
```

I want to validate that the data in the request is a valid file object (`werkzeug.datastructures.FileStorage`), but the `_deserialize` (or `_validated`) method in the `File` field is not called.

However, if I use a custom location loader for both form and files instead of `files`, then the `_deserialize` method will be called as expected:

```py
@parser.location_loader('form_and_files')
def load_form_and_files(request, schema):
    form_and_files_data = request.files.copy()
    form_and_files_data.update(request.form)
    return MultiDictProxy(form_and_files_data, schema)
```

Did I do something wrong? If it's not a bug, how to make webargs call the fields' `_deserialize` method when using `files` location?

Related to https://github.com/greyli/apiflask/pull/240

greyli:
When failed to parse a file, Werkzeug returns an empty dict (`ImmutableMultiDict([])`), then webargs will just return the `request.files` (empty dict), this makes it hard to do the further processes on the data (e.g. validation if the field contains a valid file object).

So I change this behavior to fall back on `request.form` when `request.files` is empty, this will make sure the `files` location always returns the data.

For example, I can make a custom `File` field like this:

```py
from werkzeug.datastructures import FileStorage
from marshmallow import Field


class File(Field):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.metadata['type'] = 'string'
        self.metadata['format'] = 'binary'

    default_error_messages = {
        'invalid': 'Not a valid file.'
    }

    def _deserialize(self, value, attr, data, **kwargs) -> t.Any:
        if not isinstance(value, FileStorage):
            raise self.make_error('invalid')
        return value
```

fixes #720 

If this change will be approved, I will add the related tests, docs, etc.

### DIFF
--- a/src/webargs/flaskparser.py
+++ b/src/webargs/flaskparser.py
@@ -97,7 +97,7 @@ class FlaskParser(core.Parser):
 
     def load_files(self, req, schema):
         """Return files from the request as a MultiDictProxy."""
-        return self._makeproxy(req.files, schema)
+        return self._makeproxy(req.files or req.form, schema)
 
     def handle_error(self, error, req, schema, *, error_status_code, error_headers):
         """Handles errors during parsing. Aborts the current HTTP request and


greyli:
Just realized that this solution will break the situation when `use_args` was used twice with `files` and `form`. It's more like a `form_or_files` location :P.

Closing for now.

PR #724, opened 2022-05-05
Fix DeprecationWarning for aiohttp web exceptions

### LINKED ISSUE #723: aiohttp. body argument is deprecated for http web exceptions
Hi, 
I see a lot of `DeprecationWarning: body argument is deprecated for http web exceptions` warnings.
Since aiohttp version 3.5.0, the `body` parameter has been [deprecated](https://github.com/aio-libs/aiohttp/issues/3385)

it looks like the problem is [here](https://github.com/marshmallow-code/webargs/blob/b5d76db6d7b385131a46252344b410a01bd7e852/src/webargs/aiohttpparser.py#L181)

To fix this, we just need to change the `body` to `text')

If it's OK, I can prepare a PR

AVOstap:
Resolves https://github.com/marshmallow-code/webargs/issues/723

The fix works correctly with all versions of aiohttp starting from 3.0.8

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -56,3 +56,4 @@ Contributors (chronological)
 * Legolas Bloom `@TTWShell <https://github.com/TTWShell>`_
 * Kevin Kirsche  `@kkirsche <https://github.com/kkirsche>`_
 * Isira Seneviratne `@Isira-Seneviratne <https://github.com/Isira-Seneviratne>`_
+* Anton Ostapenko `@AVOstap <https://github.com/AVOstap>`_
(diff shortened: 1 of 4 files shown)


lafrech (approved):
LGTM.

Waiting a few days in case @sirosen want to give it a look. I shall merge soon.

Thanks!

sirosen:
I've been meaning to comment on this or the issue. Everything looks good! I'll merge, and thanks for the issue+PR! 😄 

Food for thought: 3.0.8 is from 2018, which is pretty old at this point. Whenever we do 9.0, we should update our minimum supported versions for all frameworks. I'll make an issue for that.

PR #745, opened 2022-07-11
Add `flask` to mypy pre-commit dependencies

sirosen:
`flask` is needed in order to see that `flask.abort` is a `NoReturn` function. Without this, mypy incorrectly deduces that the flaskparser `abort` function has an implicit `return None`.

This is the problem that was hit during the pre-release linting for v8.2.0 . `tox -e lint` is failing. I'm not sure why it passed when it last ran in pre-commit.ci .

Longer term, I'm inclined to remove `mypy` from pre-commit and only run it in the tox env.
But for now this is a quick/easy fix.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,10 @@ repos:
   - id: mypy
     language_version: python3
     files: ^src/webargs/
-    additional_dependencies: ["marshmallow>=3,<4", "packaging"]
+    additional_dependencies:
+      - marshmallow>=3,<4
+      - packaging
+      - flask
 
 
 # mypy runs under tox in GitHub Actions, skip it in pre-commit.ci


lafrech (approved):
I'm fine with adding flask to mypy dependencies. A comment to ignore the error would be fine as well. After all it is just in one place. But I have no problem expecting developers to install flask just for mypy.

I wouldn't remove mypy from pre-commit because then mypy errors would occur in CI.

sirosen:
Right now things are simple enough that we're not seeing a ton of issues. I'm fine with whatever solution, but I think I prefer that we get the packages into the mypy env because I suspect that it will be less work overall (not fighting with false negatives).

I've also seen a trick where `pre-commit`'s `additional_dependencies` can be used to point at an extra of the current package, like `additional_dependencies: [".[tests]"]`, but I haven't tried it, so no clue how good/bad that solution is, or if it even really works.

PR #804, opened 2023-01-18
Update flake8-bugbear and black in pre-commit additional_dependencies

sirosen:
I maintain a little tool for doing this ([upadup](https://github.com/sirosen/upadup)).


lafrech (approved):
Good. Thanks for the tool.

@sloria, we might use it on other repos. Still manually triggered, but better than doing the whole thing manually.

Issue #830, opened 2023-04-25
RFC: support passing the whole parsed arguments as a keyword argument


greyli:
Hi team,

Currently, we provide two options to pass the parsed arguments to the view function:

- `use_args`
- `use_kwargs`

It seems there is a mismatch here. `use_args` passes the parsed arguments as a single positional argument, while the `use_kwargs` passes all the arguments as separate keyword arguments.

I'm thinking of adding the missing feature between this mismatch, which is to support passing the whole parsed arguments as a keyword argument.

What are the benefits? It can help to ensure the natural order for Flask views:

```py
@app.post('/pets/<pet_id>')
@use_args(PetQueryIn, location='json')
@use_args(PetBodyIn, location='query')
def create_pet(json_data, query_data, pet_id):
    pass
```

Since the URL variables (i.e. `pet_id`) will be passed as keyword arguments, we have to declare the arguments from use_args first.

Although we can use `use_kwargs` to pass all arguments as keyword arguments, but consider we will use three stacked `use_kwargs` and each schema contains five fields...

I made some experiments, there are two ways to achieve this:

- Create a new function to pass the whole parsed arguments as a keyword argument (maybe called `use_named_args`):

```py
@app.post('/pets/<pet_id>')
@use_named_args(PetQueryIn, location='json', arg_name='data')
@use_named_args(PetBodyIn, location='query', arg_name='query_data')
def create_pet(pet_id, data, query_data):
    pass
```

2. Add an `arg_name` argument to `use_args` and `use_kwargs`.

```py
@app.post('/pets/<pet_id>')
@use_kwargs(PetQueryIn, location='json', arg_name='json_data')
@use_kwargs(PetBodyIn, location='query', arg_name='query_data')
def create_pet(json_data, query_data, pet_id):
    pass
```

Do you think it's a useful feature? Which solution is preferred?

I'm happy to submit a PR. Thanks!

Related issue: https://github.com/apiflask/apiflask/issues/427

lafrech:
Yeah, current naming is not that self-explanatory.

We could change it and release a major. I'm short of better ideas, but open to suggestions.

If we keep current names, I like the second option better, keep the `use_args` name and pass an optional name.

However, unless I didn't understand the proposal, I don't think it makes sense for `use_kwargs`.

sirosen:
I agree with the premise of this issue and would like to do something about it, but I'm not sure we have my ideal solution in hand yet.

[...]

PR #833, opened 2023-04-27
Implement arg_name and USE_ARGS_POSITIONAL

sirosen:
These features allow `use_args` to pass parsed data as keyword arguments without expanding, as `as_kwargs=True`/`use_kwargs` does. This allows for better control and clarity, especially in use-cases where decorators are stacked or may be abstracted by a framework or other tooling.

``arg_name`` functions to set the keyword argument name used for parsed data (rather than appending to positional args). ``USE_ARGS_POSITIONAL = False`` on a parser class enables implicit/automatic values for ``arg_name``, making the default behavior ``{location}_args``. With this set, for example `location="json"` will pass data as `json_args`.

---

These changes and the associated docs address most of #830, but they're intentionally geared towards supporting that use case in v8 and then changing the default behavior to clean up in v9.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -681,6 +681,7 @@ of which are nearly identical:
     def viewfunc(query_args, json_args):
         ...
 
+
     # incorrect ordering, bottom-to-top
     @use_args({"foo": fields.Int(), "bar": fields.Str()}, location="query")
     @use_args({"baz": fields.Str()}, location="json")
@@ -699,11 +700,14 @@ For example,
     from webargs.flaskparser import FlaskParser
     from flask import Flask
 
+
     class KeywordOnlyParser(FlaskParser):
         USE_ARGS_POSITIONAL = False
 
+
     app = Flask(__name__)
     parser = KeywordOnlyParser()
+
 
     @app.route("/")
     @parser.use_args({"foo": fields.Int(), "bar": fields.Str()}, location="query")
@@ -718,7 +722,9 @@ parameter:
 .. code-block:: python
 
     @app.route("/")
-    @parser.use_args({"foo": fields.Int(), "bar": fields.Str()}, location="query", arg_name="query")
+    @parser.use_args(
+        {"foo": fields.Int(), "bar": fields.Str()}, location="query", arg_name="query"
+    )
     @parser.use_args({"baz": fields.Str()}, location="json", arg_name="payload")
     def myview(*, query, payload):
         ...
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -394,7 +394,6 @@ To add your own parser, extend :class:`Parser <webargs.core.Parser>` and impleme
 
     import re
 
-    from webargs import core
     from webargs.flaskparser import FlaskParser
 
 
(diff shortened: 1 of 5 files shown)


lafrech:
Thanks @sirosen for tackling this.

@greyli might want to have a look as well.

greyli:
LGTM. Thanks for the quick action!

sirosen:
I'd like to release this soon along with some more typing refinements (some of which I'm working on right now). It's been a while since our last release and I think this is a good improvement.

I've reviewed my own work with the benefit of time and space, and I'm relatively happy with the implementation.
@lafrech, if you have some time available, would you be willing to give this a final review and :+1: / :-1: it?

docs/advanced.rst
```diff
@@ -656,6 +655,83 @@ To reduce boilerplate, you could create shortcuts, like so:
...
+    ``USE_ARGS_POSITIONAL`` will be removed and will always be ``False``.
+
+By default, ``webargs`` provides two ways of passing arguments via decorators,
+`Parser.use_args <webargs.core.Parser.use_args>`, and `Parser.use_kwargs <webargs.core.Parser.use_kwargs>`.
+``use_args`` passes parsed arguments as positionals, and ``use_kwargs`` expands
+dict-like parsed arguments into keyword arguments.
+
+For ``use_args``, the result is that sometimes it is non-obvious which order
+arguments will be passed in. Consider the following two example snippets, one
+of which are nearly identical:
```
lafrech: > one of which are nearly identical

I'm not a native speaker but I find this sentence a bit strange. What about

> Consider the following nearly identical example snippets:

lafrech (commented):
Sorry about the delay, again.

Looks well thought and implemented.

What about the case of the decorator applied twice for the same location with `Schema(partial=True)`? I suppose this would be a legit case. This can work by passing explicit names but if none are passed we should probably raise an error. I didn't give it much thought but I'm not sure we can easily detect this at init.

docs/advanced.rst
sirosen: Good catch! Thanks!

I think you're right to find it strange.
"one of which are" isn't grammatical as far as I know. I'm guessing that this was an editing error of mine and that my intent was to make it read
> Consider the following two example snippets, which are nearly identical:

I prefer the phrasing you proposed, "Consider the following nearly identical example snippets", so I'll go with that.
It's a more direct construction and, IMO, easier to read.

sirosen:
> What about the case of the decorator applied twice for the same location with `Schema(partial=True)`? I suppose this would be a legit case. This can work by passing explicit names but if none are passed we should probably raise an error. I didn't give it much thought but I'm not sure we can easily detect this at init.

That's an interesting use-case. I'll see if I can do something to make this detectable.

I think it would be nice to raise an error when the decorators are evaluated, if possible. It would probably (I want to verify) raise an error at runtime.
If I can't figure it out, we could leave it as is in the current implementation but document the behavior.

sirosen:
In [41e07f1](https://github.com/marshmallow-code/webargs/pull/833/commits/41e07f13eb6592253188896653e58853bbe41eb6), I introduced detection for the name conflict case which can come from stacked decorators.

It's important to note that this starts recording information in a list, `__webargs_argnames__`, which is attached to the decorated callable. It's a new side-effect for `use_args` to have, but it follows a pattern I've seen a few other libraries use for tracking information about decorators -- I'm relatively confident that it's safe given that `click` does this. Since it's part of an opt-in feature, there isn't a worry that this could make someone's existing usage start to fail unexpectedly (e.g., if they are decorating some custom callable with `__slots__`).

lafrech:
Couldn't this be considered a breaking change in this case?

```py
        @parser.use_args(Schema_1(partial=True))
        @parser.use_args(Schema_2(partial=True))
        def viewfunc(*, j, q):
            return (j, q)
```

I believe this used to pass but will now break.

(I don't mind introducing a breaking change and moving straight to next major.)

src/webargs/core.py
```diff
@@ -51,6 +51,12 @@
 DEFAULT_VALIDATION_STATUS: int = 422
 
 
+def _record_arg_name(f: typing.Callable[..., typing.Any], argname: str) -> None:
+    if not hasattr(f, "__webargs_argnames__"):
+        f.__webargs_argnames__ = []  # type: ignore[attr-defined]
+    f.__webargs_argnames__.append(argname)  # type: ignore[attr-defined]
```
lafrech: There's a (admittedly twisted) use case I try to support in flask-smorest when stacking decorators.

Assume a view is decorated once, then used in different places with other decorators (this can be achieved by calling the decorators as normal functions, not the decorator form). Then mutating the list is wrong. It should be replaced instead.

This might not make much sense in the `use_args` case and the likelihood of someone ever doing this is close to zero but it is not forbidden...

sirosen: I had the idea yesterday that if we "just switched to a tuple" this use-case would fall out naturally from that.
When I went to do it today, I found that the only additional tweak which was needed was to put the `__webargs_argnames__` tuple on the `wrapper` built by the decorator, so I think this worked out quite well.

I've added a relevant test-case and it seems pretty good.

I wouldn't necessarily _advertise_ the feature because it's a bit unusual, but it is now supported and tested! :grin:

sirosen:
> Couldn't this be considered a breaking change in this case?
> [stacked use_args usage follows]

I think we're safe because the parser option for this controls whether or not `arg_name` is even generated inside of `use_args`, and therefore whether or not it is recorded in `__webargs_argnames__`.
Without `USE_ARGS_POSITIONAL = False` on the class, `arg_name` is `None` and `__webargs_argnames__` will never be populated.

I think we're good to release this in 8.x -- which I like better anyway, since there are other changes I'd want to include in 9.0 like removing AsyncParser from the class hierarchy and I haven't thought hard about what else we should change.

lafrech:
> I think we're safe because the parser option for this controls whether or not `arg_name` is even generated inside of `use_args`, and therefore whether or not it is recorded in `__webargs_argnames__`.
Without `USE_ARGS_POSITIONAL = False` on the class, `arg_name` is `None` and `__webargs_argnames__` will never be populated.

That's the part I was missing from the beginning.

I guess this is good to go, then. Please allow a little time for a last look and I'll merge.

Thanks.

lafrech (approved):
Great, thanks.

I'll leave you the merge/release in case you'd like to edit the commit history before merging. But It is fine as is.

sirosen:
I took care of the v8.3.0 release including this this morning. From my testing it looks good.

Issue #836, opened 2023-05-04
Please consider using Mapping instead of Dict in ArgsMap to avoid type checker errors


gh-andre:
In webargs 8.2.0, `ArgsMap` is defined as follows:
```python
ArgMap = typing.Union[
    ma.Schema,
    typing.Dict[str, typing.Union[ma.fields.Field, typing.Type[ma.fields.Field]]],
    typing.Callable[[Request], ma.Schema],
]
```
If a schema map is allocated/used this way:
```python
self.args_get = {
    "date": fields.DateTime(required = False, format="%Y-%m-%d")
}
args_form = flask_parser.parse(args_get, request, location="query")
```
, then mypy reports this error (similar for those in VS Code and VS):
> app.py:365: error: Argument 1 to "parse" of "Parser" has incompatible type "Dict[str, DateTime]"; expected "Union[Schema, Dict[str, Union[Field, Type[Field]]], Callable[[Any], Schema]]"  [arg-type]
app.py:365: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
app.py:365: note: Consider using "Mapping" instead, which is covariant in the value type

Changing the schema definition to this helps, but it must be done for all argument maps:
```python
self.args_get: Dict[str, Union[fields.Field, Type[fields.Field]]] = {
    "date": fields.DateTime(required = False, format="%Y-%m-%d")
}
```
If `ArgsMap` is changed to use `Mapping`, like this, the error is going away.
```python
ArgMap = typing.Union[
    ma.Schema,
    typing.Mapping[str, typing.Union[ma.fields.Field, typing.Type[ma.fields.Field]]],
    typing.Callable[[Request], ma.Schema],
]
```
Using Python 3.10 on Windows and mypy 1.1.1.

lafrech:
Thanks for reporting.

Without even trying to understand the error message, I'd say `Mapping` should be preferred in most cases.

Would you like to send a PR replacing `Dict` with `Mapping` in the code base?

gh-andre:
@lafrech Thanks for responding. 

> Would you like to send a PR replacing Dict with Mapping in the code base?

This may be potentially detrimental because `Dict` is mutable and `Mapping` is not, so one would need to distinguish between cases where `Mapping`, `MutableMapping` or `dict[X, Y]` when going through the entire codebase, which would probably not the best idea for someone with no experience in this code to do.

[...]

PR #847, opened 2023-06-14
Make Parser a generic type over Request

sirosen:
`webargs.core.Request` was declared as a TypeVar but then never bound to a relevant class-scope. As a result, it effectively evaluated as `Any` in all contexts.
Not only does this make typing less useful for us within webargs, it also means that users have a less clear typing contract to satisfy.

To resolve this, `Request` needs to be bound to the Parser class scope by means of the class becoming `Generic[Request]`. Subsequent to this, it becomes possible to add the relevant Request class for each framework parser as part of the type definition.

Throughout parser methods, like location loaders, every usage of the request object can be annotated with `webargs.core.Request`. This means the same thing as the relevant concrete request type for the framework parsers.

Finally, having made this change, it is possible to circle back to issue #836 and make the `ArgMap` type into a `Mapping` and have its new type propagate correctly. Note that `ArgMap` is a type alias which includes the `Request` type, meaning that it is an alias which is only valid in a context where `Request` is bound -- i.e. inside of a Parser class.

As a minor behavioral tweak required by this change, when `ArgMap` is passed to `Schema.from_dict`, it must be explicitly dict-ified if it is a non-dict Mapping. `Schema.from_dict` takes a dict input explicitly, so this is needed for types to match. New tests ensure that we behave correctly by means of UserDict.

Bump the version of mypy listed in tox config to the latest to ensure type checking runs correctly and quickly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,9 +4,19 @@ Changelog
 8.3.0 (unreleased)
 ******************
 
+Features:
+
+* ``webargs.Parser`` now inherits from ``typing.Generic`` and is parametrizable
+  over the type of the request object. Various framework-specific parsers are
+  parametrized over their relevant request object classes.
+
 Other changes:
 
-- Test against Python 3.11 (:pr:`787`).
+* Type annotations have been improved to allow ``Mapping`` for dict-like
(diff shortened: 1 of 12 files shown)


lafrech (approved):
Looks good. Thanks for tackling this!

PR #848, opened 2023-06-16
Begin disallowing untyped defs, starting with `webargs.core`

sirosen:
This is a draft, based off of #833 (mostly so that I don't need to rebase and handle conflicts). I can rebase onto `dev` if we want to merge it independently.

---

[Set mypy to disallow_untyped_defs on core.py](https://github.com/marshmallow-code/webargs/commit/da79514c4265831db7b3ad7f4b79a22ffafe1ac9)

Enforce that all function and method definitions in webargs.core have annotations for all arguments and return values.

To make it easier to roll out stricter config, move to a mypy.ini which disables the untyped-defs check on specific modules.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,39 @@
+[mypy]
+ignore_missing_imports = true
+warn_unreachable = true
+warn_unused_ignores = true
+warn_redundant_casts = true
+# warn_return_any = true
+warn_no_return = true
+no_implicit_optional = true
+disallow_untyped_defs = true
+
+[mypy-webargs.fields]
+disallow_untyped_defs = false
+
+[mypy-webargs.multidictproxy]
+disallow_untyped_defs = false
+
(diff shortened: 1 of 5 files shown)


lafrech (approved):
Thanks for handling this.

I frowned upon the need to list all modules in mypy.ini but I suppose it is meant to be temporary and ultimately all modules should comply. Perhaps we could start with a whitelist rather than a blacklist but I don't mind either way.

sirosen:
I've found that this method of listing works out well for slow migrations in the past, so I applied it here based on that success.
I agree that it's not desirable to have to list all of the modules in config, but it has some nice effects. One important detail for codebases which are moving a bit faster than `webargs` is that any new modules are included because the desired config is set as the default rule.

It should all disappear over time anyway, so I'm not too worried about the downsides of it.

PR #866, opened 2023-08-02
Fix Aiohttp ``exception_map`` type hint

grigoriev-semyon:


sirosen:
Thanks for the small fix! We'll need to setup typing tests at some point, or things like this are likely to accidentally regress.

PR #888, opened 2023-10-02
fix: schema_example.py status_code ignored

frank2411:
Just a small fix/enhancement for the examples in the webargs documentation.


sirosen:
Thanks for the contribution!

If the example is going to handle tuple return values, I think it should still call schema.dump on the result. Otherwise, the two branches are asymmetric in a potentially surprising way.

frank2411:
Hi ! Thank you for you answer !

I will adapt the code, and propose a modification

PR #892, opened 2023-10-11
Minor example improvement: support (json, status)

sirosen:
- Very minor tweak to the import order (separating stdlib from non-stdlib)
- Allow a tuple of size 2 with an int in the second slot to trigger special handling

Technically you could use a schema which supports dumping tuple values, so this really isn't any better or worse than not handling this case.
But it's just an example to show what sorts of things you can do, so this may be useful for some users.

closes #888

---

Initially, WRT #888 I was going to try to allow a lot of space for a new contributor, but it's small and I'd rather close it out by doing something proactive.
Either we agree that this is a good demo/example to add -- although it's never used by the example app -- or we close this and #888 both as "wontfix".

I don't think the change is important. But if we're going to apply it, this seems to me like a reasonable version of it.

### DIFF
--- a/examples/schema_example.py
+++ b/examples/schema_example.py
@@ -14,9 +14,9 @@ Try the following with httpie (a cURL-like utility, http://httpie.org):
     $ http GET :5001/users/ limit==1
 """
 import functools
-from flask import Flask, request
 import random
 
+from flask import Flask, request
 from marshmallow import Schema, fields, post_dump
 from webargs.flaskparser import parser, use_kwargs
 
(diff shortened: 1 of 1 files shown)


lafrech:
Thanks. Looks good enough to me.

-------------------------------------------------

To anyone interested, in flask-smorest, I crafted a util func, `unpack_tuple_response` (actually, I copied it from Flask), to parse all sorts of Flask tuple outputs.

https://github.com/marshmallow-code/flask-smorest/blob/master/flask_smorest/utils.py

PR #935, opened 2024-04-20
Fix loader functon for async methods

j0k2r:
Working with Falcon ASGI Request, webargs has issue parsing "media" inputs, because of the [`loader_func` ](https://github.com/marshmallow-code/webargs/blob/dev/src/webargs/core.py#L239) is not waited.

The code base of `iscoroutinefunction` do not detect Falcon "media" loader as coroutine function.
```python
def iscoroutinefunction(func):
    """Return True if func is a decorated coroutine function."""
    return (inspect.iscoroutinefunction(func) or
            getattr(func, '_is_coroutine', None) is _is_coroutine)
```
Getting `loader_func`  output and checking if the result is a coroutine works better.

### DIFF
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -2,10 +2,10 @@ from __future__ import annotations
 
 import asyncio
 import functools
+import inspect
 import json
 import logging
 import typing
-import inspect
 
 import marshmallow as ma
 from marshmallow import ValidationError
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -5,6 +5,7 @@ import functools
 import json
 import logging
 import typing
+import inspect
 
 import marshmallow as ma
 from marshmallow import ValidationError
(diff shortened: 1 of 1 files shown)


sirosen:
Thanks for the contribution! I think this is correct for some cases, like `req.media` for Falcon, as you say, but not for others.

If you look at the FalconParser's support for other locations (`form`, `json`, etc), you'll see that it accesses properties like `req.stream` with no notion that they may be async awaitables. I think that any framework data which can become awaitable may need additional handling, on a case-by-case basis.

Since the change is definitely not _harmful_ and works for some cases, I think we should merge and release it, and we can deal with any other async trickiness as the need arises. @lafrech, do you agree?

lafrech:
@sirosen I have no idea TBH. My knowledge about async code is limited. I don't think I'll be able to provide an educated answer any soon so I'll follow you in this one rather than block this indefinitely.
Thanks.

sirosen:
In that case, and having given myself a few more days to think about it, I'm going to merge this and write up a changelog entry for it. If I can get some time in my schedule to look into it, I might start investigating full support for async Falcon usage. If I can't, we can release with this fix on its own.

I'll run my changelog past you largely to knowledge-share how I understand this change.

PR #943, opened 2024-05-17
Fix tests for tornado+pytest issue

sirosen:
These are two small fixes to get clean test runs.

I realized that even if tornado releases with the fix from pytest, we won't instantly drop support for older tornado versions. So the workaround suggested by a pytest dev is what I'm applying here.
I setup a base test class which we can tweak appropriately and set `__test__ = False` to keep pytest from trying to run tests from that class (which fails in other ways).

On my way to this, I found I needed a very minor mypy tweak, which is isolated in a separate commit.


sloria (approved):

lafrech:
Thanks @sirosen for fixing this.

PR #944, opened 2024-05-31
Improve type annotations for the pyramid and tornado parsers

sirosen:
This changeset takes a few steps towards getting rid of mypy type checking overrides, and therefore enforcing that there are annotations everywhere.
It targets the tornado and pyramid modules.

As a related change, this updates the mypy version pinned in tox.ini to the latest.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -28,6 +28,3 @@ disallow_untyped_defs = false
 
 [mypy-webargs.falconparser]
 disallow_untyped_defs = false
-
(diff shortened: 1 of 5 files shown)


sloria (approved):

lafrech:
Thanks for taking care of this.

CI errors on 3.8. Could it be a `from future` import missing ?

sirosen:
That's exactly what it is; I'll get that fixed.

PR #960, opened 2024-07-27
Improve mypy checking on multidictproxy module

sirosen:
This change is another incremental step towards getting things passing mypy "100%" without any exemptions.
It makes some very mild runtime changes which shouldn't have any user-visible impact.

- Fix mypy to pass on the latest version
- Type annotate the MultiDictProxy class
- Add changelog note for MultiDictProxy inheritance

### DIFF
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import collections.abc
 import functools
 import inspect
 import json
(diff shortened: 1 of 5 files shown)


src/webargs/core.py
```diff
@@ -323,15 +324,19 @@ def _get_schema(self, argmap: ArgMap, req: Request) -> ma.Schema:
...
+            schema = self.schema_class.from_dict(argmap_dict)()
         elif callable(argmap):
-            schema = argmap(req)
-        elif isinstance(argmap, typing.Mapping):
-            if not isinstance(argmap, dict):
-                argmap = dict(argmap)
-            schema = self.schema_class.from_dict(argmap)()
+            # type-ignore because mypy seems to incorrectly deduce the type
+            # as `[def (Request) -> Schema] | object`
+            schema = argmap(req)  # type: ignore[call-arg, assignment]
```
lafrech, line 339: It would be nice if we had a mypy issue ID to link to. I suppose I you had found one, you would have added it to the comment. Never mind. Not a blocker.

lafrech (approved):
Thanks for the work on this.

src/webargs/core.py
sirosen: I've been meaning to circle back and try to reproduce this. I'm not aware of an open bug on `mypy`, but their issue tracker has a ton of reports and a lot of spurious/duplicate issues, so it's hard to wade through.

lafrech: Yeah, I understand. No problem. Not worth wasting too much time.
Thanks.

PR #973, opened 2024-09-10
Fix dev requirements install

un4gt:
The command `pip install -e '.[dev]'` results in a syntax error on some systems due to single quotes. 

In windows :

```
pip install -e '.[dev]'
ERROR: '.[dev]' is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).
```

By using double quotes (`pip install -e ".[dev]"`), the command runs successfully on all platforms, ensuring better cross-platform compatibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,10 +103,3 @@ venv.bak/
 
 # mypy
 .mypy_cache/
-
-# PyCharm
-#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
-#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
-#  and can be added to the global gitignore or merged into this file.  For a more nuclear
-#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,10 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# PyCharm
+#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+.idea/
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -48,7 +48,7 @@ Setting Up for Local Development
 ::
 
     # After activating your virtualenv
-    $ pip install -e '.[dev]'
+    $ pip install -e ".[dev]"
 
 3. (Optional, but recommended) Install the pre-commit hooks, which will format and lint your git staged files.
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -103,3 +103,10 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# PyCharm
+#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+.idea/
```
lafrech: This doesn't belong in the project but in [your own .gitignore](https://stackoverflow.com/a/7335487/) (local to you computer, global to your projects).

CONTRIBUTING.rst
```diff
@@ -48,7 +48,7 @@ Setting Up for Local Development
 ::
 
     # After activating your virtualenv
-    $ pip install -e '.[dev]'
+    $ pip install -e ".[dev]"
```
lafrech, line 51: It even works without any quotes on my machine (Linux). Would that work on Windows?

    pip install -e .[dev]

lafrech (changes requested):

lafrech:
The CI failure is unrelated and seems to be due to changes in bottle 0.13.

sirosen:
> The CI failure is unrelated and seems to be due to changes in bottle 0.13.

I'll plan to look into that either today or tomorrow.

lafrech:
Thanks Stephen.

It seems related to how unicode errors are treated but I haven't gotten to the bottom of it and I gotta go.

I scrolled through [bottle changelog](https://bottlepy.org/docs/dev/changelog.html#release-0-13) and nothing stroke me as relevant to this issue.

CONTRIBUTING.rst
un4gt: single quote  is notworking on windows:

image

But, double quote and without quote works well.

I think we need to unify the commands.

lafrech: Sure. I don't know why we put those quotes in the first place so I'd go for no quotes at all if it works everywhere.

sloria: oops, on zsh the quotes are required

```
$ pip install -e .[dev]                              
zsh: no matches found: .[dev]
```

double quotes seems to work though, so i'm fine with this change as is

sloria (approved):

CONTRIBUTING.rst
lafrech: OK so that's why we put those quotes in the first place. Double quotes is fine, then.

PR #974, opened 2024-09-10
Fix for invalid JSON handling on latest 'bottle'

sirosen:
bottle v0.13 captures any JSON decoding errors and reraises them as an
HTTPError exception with no inner 'exception' object and a body of
'Invalid JSON'.

Catching and handling these explicitly ensures that webargs behavior
is stable on the newer version of bottle.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ Changelog
 Unreleased
 **********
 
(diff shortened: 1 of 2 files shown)


lafrech:
Thanks!

lafrech:
Fix released. Have a nice day.

PR #981, opened 2024-10-17
Drop Python 3.8

lafrech:
Also remove Python 3.x min minor version in README and docs. Not so useful and needs to be updated on each release (which I tend to forget).

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -13,9 +13,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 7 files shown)


sirosen:
I'll put in a PR to add 3.13 as well!

lafrech:
Hold on. I'm on it already.

-- 
Jérôme

sirosen:
Holding on! Thanks for taking care of these maintenance tasks. I agree with your take that we shouldn't show the version bounds in the readme, etc.

PR #982, opened 2024-10-17
Support Python 3.13

lafrech:


lafrech:
Fails with falcon on 3.13.

No rush. Let's wait for Falcon 4.0 to ship. A RC has been published two days ago.

sirosen:
Yeah, I'm cool with that. For the record, I have a pretty good guess about the build failure.

For some reason, `setuptools` is failing to pick up the `falcon.__version__` attribute when building.
But also, I notice that their build uses Cython, so I bet it's a complex build process and not perfectly captured by the `[build-system]` table.

I'm betting that for all of the older Python versions, we've been getting wheels for `falcon`, but that there is no 3.13 wheel. That would mean this is the first time we've hit a requirement to build from sdist, and it doesn't work smoothly.

sirosen:
I was doing some personal cleanup and saw that this was still in my queue to handle. Falcon 4.0 is out, so I figured I should rerun CI and probably merge this.

Unfortunately, there's a new and different failure to handle, and we'll probably be best served waiting for this one as well.

3.13 removes `cgi` amongst the "Dead Batteries" declared in [PEP 594](https://docs.python.org/3/whatsnew/3.13.html#whatsnew313-pep594). `webob` uses `cgi`, but it looks like they're not only [working to handle it](https://github.com/Pylons/webob/pull/466) but also might ["shim" it for a bugfix release](https://github.com/Pylons/webob/pull/467).

I don't have a strong opinion about how `webob` "should" handle this, so I just look forward to picking up on their new version when it's ready. Our builds are grabbing `webob==1.8.8`, which is their latest, so I see no reason that we wouldn't be able to use the fix when it's available.

sirosen:
`webob` shipped new versions; I think we're good to go. Merging!

PR #987, opened 2025-01-07
Support marshmallow 3.24.0 and 4.0

sloria:


lafrech:
Any objection to shipping 8.7.0 now that marshmallow 4 is out?

sirosen:
None at all! :+1: I didn't realize it had already happened; I've been so out of the loop on marshmallow dev!

I'll make sure to be attentive in case we get user issue reports after the release. And I'll start to test things out myself as soon as I can.

lafrech:
8.7.0 is out!

PR #989, opened 2025-01-28
Fix the type annotation for 'argmap'

sirosen:
The ArgMap type alias defined the mapping option (for
`Schema.from_dict`) to allow `Field` or `type[Field]`, but this is
incorrect. The following construction does not work:

    Schema.from_dict({"a": fields.Str})

Correct by simply removing `type[Field]` from the alias.

A quick included improvement is to make the handling of type narrowing
for `argmap` a little bit more uniform in style (we had it 3 different
ways in different places), and to use a cast to replace a type ignore
with something a little easier to read.

### DIFF
--- a/src/webargs/core.py
+++ b/src/webargs/core.py
@@ -26,11 +26,9 @@ __all__ = [
 
 
 Request = typing.TypeVar("Request")
+ArgMapCallable = typing.Callable[[Request], ma.Schema]
 ArgMap = typing.Union[
-    ma.Schema,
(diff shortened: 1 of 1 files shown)


sloria (approved):

sirosen:
I need to look at our CI to figure out / decide how we want to handle marshmallow-dev being the 4.0 branch. But I'm comfortable ignoring that for today (don't have bandwidth to fix it right now).

Issue #725, opened 2022-05-06
Update minimum supported versions of frameworks in next major release


sirosen:
In the next major webargs release (9.0), we should make sure to update the minimum versions of the various frameworks which we test against.

IIRC, the current bounds came from looking at which versions we support in practice. But we don't want to be stuck testing against old `flask`, `django`, `pyramid`, etc versions indefinitely. We could debate exactly what kinds of updates are safe in a minor release -- especially since this is for our testing configuration and not part of our `install_requires` -- but I think it's simpler to just do this in 9.0, whenever that may be.

lafrech:
Some would say if the cost is low why not keep them, but the point is that old framework versions might not even be supported anyway so let's not try to support them here. Also the transition may be easy. I mean if someone maintains a 2018 app, I don't think they'll complain about webargs not supporting Flask 1, they'll just update Flask, the move is easy enough.

PR #997, opened 2025-04-18
Improve type annotations; avoid runtime cost

sirosen:
1. Add 'typing-extensions' as a conditional dependency. This allows us to import some of the newer symbols (at present, TypeAlias) both at runtime and at type-checking time.

2. Move type aliases into a generic '_types' helper module; make that module something we only even import under type checking.

3. Avoid all uses of `typing.cast` and rely on annotations+ type-ignore comments instead, as appropriate.

4. Fix the way that the 'Request' type var gets bound to 'ArgMap' and 'ArgMapCallable' types. (Our previous usage may have worked but was possibly relying on undefined type checker behaviors.)

Fix the (unrelated) detected "error" in the falcon parser.

Overall, this should take type annotations for `webargs` to a nearly 0 runtime cost. But it also improves the accuracy and maintainability of the annotations.

---

I realized that something was sort of wrong with the `ArgMap` types.
I wanted to wrangle with that and figure out how to check more accurately.
Trying to remove runtime costs was a small add-on while looking at our typing usages.

I tested using this sample to confirm that this produces the correct type annotation for `req`[^1]:

```python
# sample.py
import webargs

class MyReqType:
    pass

class MyParser(webargs.core.Parser[MyReqType]):
    pass

result = MyParser().parse({}, req=object())
```

```bash
$ mypy sample.py
sample.py:10: error: Argument "req" to "parse" of "Parser" has incompatible type "object"; expected "MyReqType | None"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

[^1]: This already works today against `dev`, so this is a regression test. But I'm not clear on whether or not our usage might stop working as typing gets more standardized, since we're a bit off-spec.

---

I re-checked and got a test case which demonstrates the improvement.

```python
# sample2.py
import marshmallow
import webargs

class MyReqType:
    pass

class MyParser(webargs.core.Parser[MyReqType]):
    pass

def dynamic_schema(foo: int) -> marshmallow.Schema:
    return marshmallow.Schema.from_dict({})()

result = MyParser().parse(dynamic_schema)
```

On `dev`, `mypy` passes on this, no errors.
With these changes:
```bash
$ mypy sample2.py       
sample2.py:14: error: Argument 1 to "parse" of "Parser" has incompatible type "Callable[[int], Schema]"; expected "Schema | type[Schema] | Mapping[str, Field[Any]] | Callable[[MyReqType], Schema]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

Because it's able to recognize that the callable which is supposed to be of type `MyReqType -> Schema` is of type `int -> Schema`. On `dev`, the issue is that the type of that callable is written as `Request -> Schema`, but `Request` isn't properly bound, so it's _really_ `Any -> Schema`.

The typing stuff can still get me a little turned around sometimes.

### DIFF
--- a/src/webargs/falconparser.py
+++ b/src/webargs/falconparser.py
@@ -38,19 +38,18 @@ def parse_form_body(req: falcon.Request):
         and "application/x-www-form-urlencoded" in req.content_type
     ):
         # the type of `req.stream.read()` is `str | Any` according to annotations
-        # but at runtime it's always `bytes`, so coerce by checking the type
+        # but at runtime it's always `bytes`, so coerce the type for type checker
         body: str | bytes | None = req.stream.read(req.content_length or 0)
-        if isinstance(body, bytes):
-            try:
-                body = body.decode("ascii")
-            except UnicodeDecodeError:
-                body = None
-                req.log_error(
-                    "Non-ASCII characters found in form body "
-                    "with Content-Type of "
-                    "application/x-www-form-urlencoded. Body "
-                    "will be ignored."
-                )
+        try:
+            body = body.decode("ascii")  # type: ignore[union-attr]
+        except UnicodeDecodeError:
+            body = None
+            req.log_error(
+                "Non-ASCII characters found in form body "
+                "with Content-Type of "
+                "application/x-www-form-urlencoded. Body "
+                "will be ignored."
+            )
 
         if body:
             return parse_query_string(body, keep_blank=req.options.keep_blank_qs_values)
--- a/src/webargs/falconparser.py
+++ b/src/webargs/falconparser.py
@@ -1,5 +1,7 @@
 """Falcon request argument parsing module."""
 
+from __future__ import annotations
+
 import falcon
 import marshmallow as ma
 from falcon.util.uri import parse_query_string
@@ -35,17 +37,20 @@ def parse_form_body(req: falcon.Request):
         req.content_type is not None
         and "application/x-www-form-urlencoded" in req.content_type
     ):
-        body = req.stream.read(req.content_length or 0)
-        try:
-            body = body.decode("ascii")
-        except UnicodeDecodeError:
-            body = None
-            req.log_error(
-                "Non-ASCII characters found in form body "
-                "with Content-Type of "
-                "application/x-www-form-urlencoded. Body "
-                "will be ignored."
-            )
+        # the type of `req.stream.read()` is `str | Any` according to annotations
+        # but at runtime it's always `bytes`, so coerce by checking the type
+        body: str | bytes | None = req.stream.read(req.content_length or 0)
+        if isinstance(body, bytes):
+            try:
+                body = body.decode("ascii")
+            except UnicodeDecodeError:
+                body = None
+                req.log_error(
+                    "Non-ASCII characters found in form body "
+                    "with Content-Type of "
+                    "application/x-www-form-urlencoded. Body "
+                    "will be ignored."
+                )
 
         if body:
             return parse_query_string(body, keep_blank=req.options.keep_blank_qs_values)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,12 @@ keywords = [
   "marshmallow",
 ]
 requires-python = ">=3.9"
-dependencies = ["marshmallow>=3.0.0", "packaging>=17.0"]
+dependencies = [
+    "marshmallow>=3.0.0",
+    "packaging>=17.0",
+    # depend on typing-extensions conditionally for annotations
+    'typing_extensions>=4.0; python_version<"3.10"',
+]
 
 [project.urls]
 Changelog = "https://webargs.readthedocs.io/en/latest/changelog.html"
--- a/src/webargs/_types.py
+++ b/src/webargs/_types.py
@@ -1,0 +1,34 @@
+from __future__ import annotations
+
+import sys
+import typing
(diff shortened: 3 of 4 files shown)


src/webargs/falconparser.py
```diff
@@ -35,17 +37,20 @@ def parse_form_body(req: falcon.Request):
...
-            req.log_error(
-                "Non-ASCII characters found in form body "
-                "with Content-Type of "
-                "application/x-www-form-urlencoded. Body "
-                "will be ignored."
-            )
+        # the type of `req.stream.read()` is `str | Any` according to annotations
+        # but at runtime it's always `bytes`, so coerce by checking the type
+        body: str | bytes | None = req.stream.read(req.content_length or 0)
+        if isinstance(body, bytes):
```
lafrech: I don't understand why this is better than e.g. a cast or and ignore.

lafrech (commented):
Thanks.

Looks good to me and well motivated. Honestly I'm trusting your on the complicated parts.

I can't comment on good practices but I find it nice to group stuff in a dedicated file.

My concern with advance typing in general is how it can make the codebase more complicated to work on. I'm thinking of those times where I want to make a simple change and typing gets in the way. I guess I gotta get used to it and learn.

sloria (approved):
Nice changes! Only skimmed but nothing looks obviously incorrect. No need to block on me for merging

src/webargs/falconparser.py
sirosen: Let me double-check this before we consider merging. At the time that I was touching this, I was under the impression that this `read()` could sometimes return a `str`, but now I'm no longer sure.

sirosen: Let me turn this into a type-ignore.

I have tried a few specific things against the latest `falcon` version, and I can't make `req.stream.read` return anything other than bytes.

Thanks for calling this out!

sirosen:
> I can't comment on good practices but I find it nice to group stuff in a dedicated file.

Yeah, that was a big part of my thinking here. I'm not placing too much importance on deferring this stuff, but it felt like some special typing stuff should have a proper home.

> My concern with advance typing in general is how it can make the codebase more complicated to work on. I'm thinking of those times where I want to make a simple change and typing gets in the way. I guess I gotta get used to it and learn.

Yep, 100% a valid concern, and one which I share! We have some codebases at work with some fancy typing usage and it's more complex than I would like. I'm always available if you have questions where it impacts webargs or other marshmallow projects.

sirosen:
I'm going ahead with the merge. I'm always open to tuning this (or even scaling back annotations) more if necessary.

sloria:
thanks!

src/webargs/falconparser.py
vytas7: Hi there :wave:

Could you elaborate on when reading [`req.stream`](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.stream) yields `str` or `None`?

This is not something we statically declare either:
```python
import typing

import falcon
import falcon.testing


def main() -> None:
    req = falcon.testing.create_req()
    typing.reveal_type(req.stream.read)


if __name__ == '__main__':
    main()
```
:arrow_down: 
```
$ mypy --strict test.py
test.py:9: note: Revealed type is "def (builtins.int | None =) -> builtins.bytes"
Success: no issues found in 1 source file
```

sirosen: Hi, thanks for reaching out to a downstream project!

I recall with some clarity that I looked at the `falcon` source at the time and couldn't find an obvious explanation, but that `mypy` was flagging this type as `bytes | str | Any` when I was running it. It didn't make a ton of sense to me, but I couldn't quickly rule out the possibility that I was not understanding the `falcon` source.

I just tried checking things out with stuff pinned back to `falcon<4.1`, `mypy==1.11.0`, and even `typing_extensions==4.0`, and I can't reproduce it. So I don't think there's an upstream issue at all!

We could update this block of code to drop the try-decode step, but I'd want to check that against very old `falcon` versions too.

vytas7: You're welcome!

Hm, reading from the input stream has always been `bytes`, I don't think you need to check anything.

One possible reason could be that _webargs_ still support Falcon 2.0, which in turn still supported Python 2.7 back in the day where `str` and `bytes` were synonymous (instead you could explicitly use `unicode` strings). Maybe that's where the confusion is stemming from? 
However, if your project no longer supports Python 2.7, you can safely ignore this possibility altogether.

Otherwise you can safely switch to just `bytes` in #725.

vytas7: P.S. Note that if you do drop Falcon 2.0 support in #725, you can rewrite the whole `application/x-www-form-urlencoded` block simply as [`req.get_media()`](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_media):
```python
>>> import falcon
>>> import falcon.testing
>>> 
>>> class MediaPrinter:
...     def on_post(self, req, resp):
...         print(req.content_type, req.get_media())
... 
>>> app = falcon.App()
>>> app.add_route('/', MediaPrinter())
>>>
>>> falcon.testing.simulate_post(app, '/', body='color=green&project=webargs', content_type='application/x-www-form-urlencoded')
application/x-www-form-urlencoded {'color': 'green', 'project': 'webargs'}
Result<200 OK application/json b''>
```

PR #1019, opened 2025-10-29
Drop Python 3.9

lafrech:
EOF in two days: https://endoflife.date/python

I had to bump bottle and tornado versions to support Python 3.10.

In marshmallow, we don't consider dropping EOL Python versions a breaking change. Since here it implies dropping framework versions, we may want to release a webargs 9.0 and bump framework versions even further (#725).


sloria:
yeah, agreed on bumping up minimum framework version support

PR #1023, opened 2025-11-02
Bump supported framework versions

### LINKED ISSUE #725: Update minimum supported versions of frameworks in next major release

lafrech:
Closes #725.

I bumped to the latest minor, except for Pyramid for which it was pretty old and Python 3.10 and 3.11 support was added in subsequent patch release.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -13,10 +13,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { name: "3.10", python: "3.10", tox: py310-marshmallow3 }
-          - { name: "3.14", python: "3.14", tox: py314-marshmallow3 }
-          - { name: "3.10", python: "3.10", tox: py310-marshmallow4 }
-          - { name: "3.14", python: "3.14", tox: py314-marshmallow4 }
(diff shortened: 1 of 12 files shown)


sloria (approved):

lafrech:
Alright. Sorry about the push force spam. I missed a few code cleanups at first.

I also bumped marshmallow to v4. Is that fine?

sloria:
good idea!

sirosen (approved):
Thanks for taking care of this!

I'm good with bumping to marshmallow 4+ , as part of the general suite of moves for v9.

I'd like a chance to look around and see if there's anything more to clean up before we release. Would it help to plan a rough date for release, maybe at the end of the month?

PR #1024, opened 2025-11-02
Remove __version__ et al

lafrech:
We didn't deprecate in 8.x but we may remove in a major version anyway.

Or would you like to deprecate in 9.0 first?


sirosen (approved):
I'm happy for us to remove this now, without a deprecation period.

Users can update to `importlib.metadata` and they'll be fine. Enough packages have already been trending in this direction, I bet that most webargs users are already exposed to the newer usage.

PR #1033, opened 2026-01-30
Remove the separate async parser class for v9.0

sirosen:
This took a while for me to open as a PR, but it's been on my radar for a long time as a change I wanted to make for the next major version.
Because webargs has been so stable, it hasn't been a priority.

@lafrech, @sloria, I'm open to discussing it more -- I care more about the conceptual cleanup than any maintenance burden we carry.
`AsyncParser` is very small and doesn't cost much to keep.
It could remain but raise a deprecation warning on init in v9.x , and we'd wait for our next major to eventually drop it.

---

The separate AsyncParser class is no longer needed, as async parsing is
now handled by the main Parser class. This transition was achieved during
8.x versions and therefore AsyncParser had to remain as a shim which
replaces `parse` (the sync function on the base class) with a call to the
async variant, `async_parse`.

A battery of tests with called `await parser.parse(...)` on an async
parser are trivially updated to `await parser.async_parse(...)`, but no
other changes are needed.


lafrech (approved):
I agree with the intent. Thanks for the cleanup.
I'm happy doing this in v9.
Not sure it is worth the deprecation notice but if we want to do that we can deprecate in v8.x.

sloria (approved):

sirosen:
> Not sure it is worth the deprecation notice but if we want to do that we can deprecate in v8.x.

I think let's not bother. Most users would be unaffected, and it's possible to swap `parse` for `async_parse` (like I did in the testsuite) even on v8.x.

PR #1037, opened 2026-03-02
Temporarily pin `setuptools<82.0` for pyramid tests

sirosen:
`pyramid` uses `pkg_resources`, so we need a pin, at least for now.

It's done in `tox.ini` to be very lightweight.

---

I want to also try switching the test configuration so that we have a matrix
where each framework is independent, and all of them are declared as dependency
groups.

I think this will require adjustments to make the per-framework tests
conditional on their relevant framework being installed.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,6 +19,12 @@ deps =
     lowest: falcon==4.1.0
     lowest: aiohttp==3.13.0
     lowest: marshmallow==3.13.0
+
+    # temporary pin: hold back `setuptools` so that `pyramid` testing works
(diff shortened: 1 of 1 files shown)


lafrech (approved):
pyramid [2.1a1](https://github.com/Pylons/pyramid/releases/tag/2.1a1) adds that pin in their pyproject.toml.

But I don't know when 2.1 ships and I'm fine adding this here until this is dealt with properly there.

It's not as if it conflicts with anything else. And it's only test setup.

PR #1039, opened 2026-03-03
Convert optional-dependencies to dependency-groups

sirosen:
I started this work with the thought that there's been a long-standing maintenance risk for `webargs` that I wanted to mitigate.
If our frameworks ever become incompatible with one another, we won't be able to install them all at once for doc builds or testing.

However, after more thought and evaluation, actually *doing* the testing isolated per framework would make CI slower and less energy efficient, for unclear gain.
Instead of making the change to test each framework independently, this changeset prepares for the possibility of doing the testing separately, but without acting on it.

For docs, however, it is a little bit worthwhile to use sphinx autodoc's import-mock feature to remove the dependency on frameworks.
Not only does this make our doc builds keep working if the frameworks ever cause issues, it also makes doc builds more efficient.

As a bonus inclusion, in order to make RTD builds use the newly defined `docs` dependency group, convert to a `uv`-based doc build which reuses our tox environment.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,12 +4,15 @@ sphinx:
 formats:
   - pdf
 build:
-  os: ubuntu-22.04
-  tools:
-    python: "3.11"
-python:
-  install:
-    - method: pip
-      path: .
-      extra_requirements:
-        - docs
+  os: ubuntu-24.04
+  commands:
+    # setup uv
+    - asdf plugin add uv
+    - asdf install uv latest
+    - asdf global uv latest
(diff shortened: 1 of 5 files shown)


sloria (approved):
very nice. we should probably do this across all marshmallow repos, too

sirosen:
I'm happy to put in the work to do it!

dependency-groups are my pet project, after all. :grin:

sloria:
Really appreciate it ❣️

lafrech:
I had never heard about dependency-groups.

Congratulations for that work.

PR #1078, opened 2026-07-18
Fix PyramidParser.use_args ignoring instance location and per-location unknown defaults

ArmaanjeetSandhu:
`PyramidParser.use_args` fully overrides the base `Parser.use_args`, but its signature defaults drifted out of sync with the base class and were never updated when per-location `unknown` handling was introduced (commit 8278374). As a result, calling `PyramidParser.use_args(...)` directly (without passing `location`/`unknown`) behaves differently from `parse()`, from `use_kwargs()`, and from every other framework parser.

This PR realigns the two defaults with the base class.

## Bugs fixed

1. **`DEFAULT_UNKNOWN_BY_LOCATION` was bypassed.** The override defaulted to `unknown=None` ("pass nothing, use the schema's own setting") instead of the `_UNKNOWN_DEFAULT_PARAM` sentinel that engages the per-location defaults. So `@parser.use_args(schema, location="querystring")` did not `EXCLUDE` unknown fields, and `matchdict`/`path` did not `RAISE` as intended.
2. **Instance-level `location` was ignored.** The override defaulted to `location=core.Parser.DEFAULT_LOCATION` (`"json"`, truthy), so `location = location or self.location` never fell through to `self.location`. A `PyramidParser(location="querystring")` instance still defaulted `use_args` to `json`.

Both defects only manifested when calling `use_args` directly; `use_kwargs` and `parse()` were unaffected because they pass explicit sentinel/`None` values.

## Important note

This restores the documented/intended behavior, but is technically a behavior change for anyone who was inadvertently relying on the buggy defaults (e.g. getting `RAISE` on querystrings via `use_args`). Flagging in case it warrants a changelog entry under bug fixes.


sloria:
hello. have you taken the time to review our AI policy?

sirosen:
By sheer serendipity, I'm looking at this at exactly the same time!
I think there is more than enough signal to close this as unhelpful.

@ArmaanjeetSandhu, if you can *succinctly* describe something we should fix, please open an issue first. You are still very welcome to join us and help, but only if you are willing to put in the necessary work.

---

@sloria, if you ever want to sync up and talk about how we handle such cases, feel free to ping me. I've been doing an increasing amount of work in this space.

sloria:
@sirosen curious to hear what you've been working on. feel free to email me

